PR #3, opened 2022-03-05
8253424: Add support for running pre-submit testing using GitHub Actions

gnu-andrew:
Initial attempt to backport https://github.com/openjdk/jdk11u-dev/commit/1faefed218051c324bdb5c7c10369050d6c9dd44

Changes so far:
* `make/autoconf/version-numbers` -> `common/autoconf/version-numbers`
* Fix version variables (`JDK_MAJOR_VERSION`, `JDK_MINOR_VERSION` and `JDK_MICRO_VERSION` in 8u)
* Remove boot JDK for Linux (provided by build environment Ubuntu 20.04 in openjdk-8-jdk)
* Replace `--with-version-opt` with `--with-user-release-suffix`
* Drop `--enable-jtreg-failure-handler` (unsupported)
* Move make targets to make invocation (`--with-default-make-target` not supported)
* Drop `--with-msvc-toolset-version`
* Disable tests as we don't currently have the bundles required to support them

---------
### Progress
- [x] Change must be properly reviewed

### Issues
 * [JDK-8253424](https://bugs.openjdk.java.net/browse/JDK-8253424): Add support for running pre-submit testing using GitHub Actions
 * [JDK-8253865](https://bugs.openjdk.java.net/browse/JDK-8253865): Pre-submit testing using GitHub Actions does not detect failures reliably
 * [JDK-8254054](https://bugs.openjdk.java.net/browse/JDK-8254054): Pre-submit testing using GitHub Actions should not use the deprecated set-env command
 * [JDK-8254173](https://bugs.openjdk.java.net/browse/JDK-8254173): Add Zero, Minimal hotspot targets to submit workflow
 * [JDK-8254175](https://bugs.openjdk.java.net/browse/JDK-8254175): Build no-pch configuration in debug mode for submit checks
 * [JDK-8254282](https://bugs.openjdk.java.net/browse/JDK-8254282): Add Linux x86_32 builds to submit workflow
 * [JDK-8255373](https://bugs.openjdk.java.net/browse/JDK-8255373): Submit workflow artifact name is always "test-results_.zip"
 * [JDK-8255895](https://bugs.openjdk.java.net/browse/JDK-8255895): Submit workflow artifacts miss hs_errs/replays due to ZIP include mismatch
 * [JDK-8256127](https://bugs.openjdk.java.net/browse/JDK-8256127): Add cross-compiled foreign architectures builds to submit workflow
 * [JDK-8256277](https://bugs.openjdk.java.net/browse/JDK-8256277): Github Action build on macOS should define OS and Xcode versions
 * [JDK-8256354](https://bugs.openjdk.java.net/browse/JDK-8256354): Github Action build on Windows should define OS and MSVC versions
 * [JDK-8256414](https://bugs.openjdk.java.net/browse/JDK-8256414): add optimized build to submit workflow
 * [JDK-8256393](https://bugs.openjdk.java.net/browse/JDK-8256393): Github Actions build on Linux should define OS and GCC versions
 * [JDK-8256747](https://bugs.openjdk.java.net/browse/JDK-8256747): GitHub Actions: decouple the hotspot build-only jobs from Linux x64 testing
 * [JDK-8257056](https://bugs.openjdk.java.net/browse/JDK-8257056): Submit workflow should apt-get update to avoid package installation errors
 * [JDK-8259924](https://bugs.openjdk.java.net/browse/JDK-8259924): GitHub actions fail on Linux x86_32 with "Could not configure libc6:i386"
 * [JDK-8260460](https://bugs.openjdk.java.net/browse/JDK-8260460): GitHub actions still fail on Linux x86_32 with "Could not configure libc6:i386"
 * [JDK-8263667](https://bugs.openjdk.java.net/browse/JDK-8263667): Avoid running GitHub actions on branches named pr/*
 * [JDK-8255305](https://bugs.openjdk.java.net/browse/JDK-8255305): Add Linux x86_32 tier1 to submit workflow
 * [JDK-8255352](https://bugs.openjdk.java.net/browse/JDK-8255352): Archive important test outputs in submit workflow
 * [JDK-8282225](https://bugs.openjdk.java.net/browse/JDK-8282225): GHA: Allow one concurrent run per PR only

### Reviewers
 * [Severin Gehwolf](https://openjdk.java.net/census#sgehwolf) (@jerboaa - **Reviewer**)

### Contributors
 * Alex Kasko `<akasko@openjdk.org>`
 * Zdeněk Žamberský `<zzambers@redhat.com>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk8u-dev pull/3/head:pull/3` \
`$ git checkout pull/3`

`$ git checkout pull/3` \
`$ git pull https://git.openjdk.java.net/jdk8u-dev pull/3/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 3`

`$ git pr show -t 3`

</details>

<a href="https://git.openjdk.java.net/jdk8u-dev/pull/3.diff">https://git.openjdk.java.net/jdk8u-dev/pull/3.diff</a>

</details>

### DIFF
--- a/make/conf/test-dependencies
+++ b/make/conf/test-dependencies
@@ -29,9 +29,22 @@ BOOT_JDK_VERSION=8
 JTREG_VERSION=5.1
 JTREG_BUILD=b01
 
+VS2010_FILENAME=VS2010Express1.iso
+VS2010_DIR=VS2010Express1
+VS2010_URL=https://web.archive.org/web/20140227220734if_/http://download.microsoft.com/download/1/E/5/1E5F1C0A-0D5B-426A-A603-1798B951DDAE/VS2010Express1.iso
+VS2010_SHA256=a9d5dcdf55e539a06547a8ebbc63d55dc167113e09ee9e42096ab9098313039b
+
+VS2017_FILENAME=vs2017.exe
+VS2017_URL=https://github.com/akashche/msvs_2017_installer_bootstrap/raw/master/vs_community__7955ddbf8a9b49dda0f8d18876e93bd2.exe
+VS2017_SHA256=ca4ebbc3bee27caa44ef91b21bc30837a028c23a046c009600038316e5976b5b
+
 WINDOWS_X64_BOOT_JDK_FILENAME=openjdk-8_windows-x64_bin.zip
 WINDOWS_X64_BOOT_JDK_URL=https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u322-b06/OpenJDK8U-jdk_x64_windows_hotspot_8u322b06.zip
 WINDOWS_X64_BOOT_JDK_SHA256=c9e06afb5df850e90a4da9da31c2edf10bd6da9962c4b253e91b41237f8fb2fb
+
+WINDOWS_X86_BOOT_JDK_FILENAME=openjdk-8_windows-x32_bin.zip
+WINDOWS_X86_BOOT_JDK_URL=https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u322-b06/OpenJDK8U-jdk_x86-32_windows_hotspot_8u322b06.zip
+WINDOWS_X86_BOOT_JDK_SHA256=ec96bf3ff707e91924ffe09227dcab85d78bfd565dd85b9c4eace41a7d5f4e6e
 
 MACOS_X64_BOOT_JDK_FILENAME=openjdk-8_osx-x64_bin.tar.gz
 MACOS_X64_BOOT_JDK_URL=https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u322-b06/OpenJDK8U-jdk_x64_mac_hotspot_8u322b06.tar.gz
--- a/make/conf/test-dependencies
+++ b/make/conf/test-dependencies
@@ -1,0 +1,38 @@
+#
+# Copyright (c) 2020, 2021, Oracle and/or its affiliates. All rights reserved.
+# DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+#
+# This code is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 only, as
+# published by the Free Software Foundation.  Oracle designates this
+# particular file as subject to the "Classpath" exception as provided
+# by Oracle in the LICENSE file that accompanied this code.
+#
+# This code is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+# version 2 for more details (a copy is included in the LICENSE file that
+# accompanied this code).
+#
+# You should have received a copy of the GNU General Public License version
+# 2 along with this work; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+# or visit www.oracle.com if you need additional information or have any
+# questions.
+#
+
+# Versions and download locations for dependencies used by pre-submit testing.
+
+BOOT_JDK_VERSION=8
+JTREG_VERSION=5.1
+JTREG_BUILD=b01
+
+WINDOWS_X64_BOOT_JDK_FILENAME=openjdk-11_windows-x64_bin.zip
+WINDOWS_X64_BOOT_JDK_URL=https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u322-b06/OpenJDK8U-jdk_x64_windows_hotspot_8u322b06.zip
+WINDOWS_X64_BOOT_JDK_SHA256=c9e06afb5df850e90a4da9da31c2edf10bd6da9962c4b253e91b41237f8fb2fb
+
+MACOS_X64_BOOT_JDK_FILENAME=openjdk-11_osx-x64_bin.tar.gz
+MACOS_X64_BOOT_JDK_URL=https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u322-b06/OpenJDK8U-jdk_x64_mac_hotspot_8u322b06.tar.gz
+MACOS_X64_BOOT_JDK_SHA256=96a3124bf0f5ca777954239893cc89ea34c4bc9a9b7c1559aa2c69baa0ee84e3
--- a/make/conf/test-dependencies
+++ b/make/conf/test-dependencies
@@ -29,10 +29,10 @@ BOOT_JDK_VERSION=8
 JTREG_VERSION=5.1
 JTREG_BUILD=b01
 
-WINDOWS_X64_BOOT_JDK_FILENAME=openjdk-11_windows-x64_bin.zip
+WINDOWS_X64_BOOT_JDK_FILENAME=openjdk-8_windows-x64_bin.zip
 WINDOWS_X64_BOOT_JDK_URL=https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u322-b06/OpenJDK8U-jdk_x64_windows_hotspot_8u322b06.zip
 WINDOWS_X64_BOOT_JDK_SHA256=c9e06afb5df850e90a4da9da31c2edf10bd6da9962c4b253e91b41237f8fb2fb
 
-MACOS_X64_BOOT_JDK_FILENAME=openjdk-11_osx-x64_bin.tar.gz
+MACOS_X64_BOOT_JDK_FILENAME=openjdk-8_osx-x64_bin.tar.gz
 MACOS_X64_BOOT_JDK_URL=https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u322-b06/OpenJDK8U-jdk_x64_mac_hotspot_8u322b06.tar.gz
 MACOS_X64_BOOT_JDK_SHA256=96a3124bf0f5ca777954239893cc89ea34c4bc9a9b7c1559aa2c69baa0ee84e3
--- a/.github/workflows/freetype.vcxproj
+++ b/.github/workflows/freetype.vcxproj
@@ -1,0 +1,1733 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Debug|Win32">
+      <Configuration>Debug</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|x64">
+      <Configuration>Debug</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug Multithreaded|Win32">
+      <Configuration>Debug Multithreaded</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug Multithreaded|x64">
+      <Configuration>Debug Multithreaded</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug Singlethreaded|Win32">
+      <Configuration>Debug Singlethreaded</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug Singlethreaded|x64">
+      <Configuration>Debug Singlethreaded</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|Win32">
+      <Configuration>Release</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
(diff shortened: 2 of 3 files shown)


jerboaa:
@gnu-andrew Could you mark this PR as draft until it's ready for review, please? This avoids many emails sent to the list unnecessarily.

gnu-andrew:
> @gnu-andrew Could you mark this PR as draft until it's ready for review, please? This avoids many emails sent to the list unnecessarily.

Done. Sorry, I didn't realise it wasn't spamming anyone.

I've done as much as I can here. I'm going to need to enlist someone to look at the Windows build. The additional builds & tests will need backports of bundle support.

jerboaa:
> I've done as much as I can here. I'm going to need to enlist someone to look at the Windows build. The additional builds & tests will need backports of bundle support.

Maybe @akashche  or @stooke can help with the Windows GHA builds?

akashche:
@jerboaa , I don't have experience with GitHub Actions, will look into them and report back.

gdams:
> Maybe @akashche or @stooke can help with the Windows GHA builds?

I've played around with trying to get JDK8u to compile on Windows in GH actions before, I think it mainly boils down to the version of Visual Studio on the executors. At Eclipse Adoptium we use Visual Studio 2013 but there is no easy way to install that onto the environment anymore (we used to use chocolatey but that package broke a while back). Happy to provide any help if required

gnu-andrew:
> > Maybe @akashche or @stooke can help with the Windows GHA builds?
> 
> I've played around with trying to get JDK8u to compile on Windows in GH actions before, I think it mainly boils down to the version of Visual Studio on the executors. At Eclipse Adoptium we use Visual Studio 2013 but there is no easy way to install that onto the environment anymore (we used to use chocolatey but that package broke a while back). Happy to provide any help if required

Any help is very much welcome. I've never built on Windows before, so I have no idea of the requirements and issues there. What we have at present is pretty much what is in 11u-dev, other than I removed the `--with-msvc-toolset-version` option which 8u doesn't support. Maybe we have to either look into getting 8u to support newer versions of Windows or forego automated Windows testing. The former is risky, but preferable to the latter.

make/conf/test-dependencies
```diff
@@ -0,0 +1,38 @@
...
+# questions.
+#
+
+# Versions and download locations for dependencies used by pre-submit testing.
+
+BOOT_JDK_VERSION=8
+JTREG_VERSION=5.1
+JTREG_BUILD=b01
+
+WINDOWS_X64_BOOT_JDK_FILENAME=openjdk-11_windows-x64_bin.zip
```
gdams, line 32:
```suggestion
WINDOWS_X64_BOOT_JDK_FILENAME=openjdk-8_windows-x64_bin.zip
```

make/conf/test-dependencies
```diff
@@ -0,0 +1,38 @@
...
+WINDOWS_X64_BOOT_JDK_URL=https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u322-b06/OpenJDK8U-jdk_x64_windows_hotspot_8u322b06.zip
+WINDOWS_X64_BOOT_JDK_SHA256=c9e06afb5df850e90a4da9da31c2edf10bd6da9962c4b253e91b41237f8fb2fb
+
+MACOS_X64_BOOT_JDK_FILENAME=openjdk-11_osx-x64_bin.tar.gz
```
gdams, line 36:
```suggestion
MACOS_X64_BOOT_JDK_FILENAME=openjdk-8_osx-x64_bin.tar.gz
```

zzambers:
I was doing some experiments with windows (32-bit) builds of jdk8 in github actions in the past using vs2010express.  (Visual Studio 2010 is mentioned for windows in README-builds.html) When I tried my old code, and have found, that it no longer works (simple steps to install vs2010express no longer work, caused by dead MS links [1]). However, after some experimentation, I was able to install vs2010express from iso (from webarchive) and now it works again.

My basic (cleaned up) github workflow which does 32-bit build of openjdk8 using vs2010express can be found here. [2]
(feel free to use any of that code)

[1] https://github.community/t/how-to-use-vc100-for-msbuild/16623/2
[2] https://github.com/zzambers/test-jdk-builder/blob/build-ojdk8-win/.github/workflows/build-workflow.yaml

zzambers:
But there is one problem, which needs to workarounded (see that ugly sed [3] in my workflow) in my setup, to get through configure as otherwise it dies with this error (seems like problem with path conversion somewhere, but I don't remember exactly as this is from my old code):

```
...
checking if fixpath can be created... no
Microsoft (R) 32-bit C/C++ Optimizing Compiler Version 16.00.30319.01 for 80x86
Copyright (C) Microsoft Corporation.  All rights reserved.

fixpath.c
Microsoft (R) Incremental Linker Version 10.00.30319.01
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:D:A:/test-jdk-builder/test-jdk-builder/jdk8u-dev/build/windows-x86-normal-server-fastdebug/fixpath.exe 
fixpath.obj 
LINK : fatal error LNK1104: cannot open file 'D:A:/test-jdk-builder/test-jdk-builder/jdk8u-dev/build/windows-x86-normal-server-fastdebug/fixpath.exe'
configure: error: Could not create /D/a/test-jdk-builder/test-jdk-builder/jdk8u-dev/build/windows-x86-normal-server-fastdebug/fixpath.exe
```

[3] https://github.com/zzambers/test-jdk-builder/blob/ff5f0c9662dc6829c1b267cecfdfdc74ed0af3a1/.github/workflows/build-workflow.yaml#L58

shipilev:
@gnu-andrew -- please consider adding [JDK-8282225](https://bugs.openjdk.java.net/browse/JDK-8282225) to this batch as well.

gdams:
@zzambers those runs look promising, can you try with VS 2013?

jerboaa:
> Pre-submit tests - Linux additional - Build Failing after 24s — 8/8 failed 

Should we disable those cross compile builds for the initial integration?

make/conf/test-dependencies
gnu-andrew: Thanks for catching this. It's only the name for the downloaded file, but it's better that the label matches what's in the box :)

gnu-andrew:
> > Pre-submit tests - Linux additional - Build Failing after 24s — 8/8 failed
> 
> Should we disable those cross compile builds for the initial integration?

Yes, I actually already tried to do that but it's not easy to do in a non-destructive way. I'll try again with this next commit.

gnu-andrew:
> > > Pre-submit tests - Linux additional - Build Failing after 24s — 8/8 failed
> > 
> > 
> > Should we disable those cross compile builds for the initial integration?
> 
> Yes, I actually already tried to do that but it's not easy to do in a non-destructive way. I'll try again with this next commit.

For the record, the issue is the same as with the tests; bundles. It needs the host JDK built by the x86_64 build for the cross-compiled builds, and that is cached and then retrieved via a bundle. I'll open a separate PR to backport bundles after this. That should fix additional HS builds, but tests probably need more backporting work.

gnu-andrew:
> @gnu-andrew -- please consider adding [JDK-8282225](https://bugs.openjdk.java.net/browse/JDK-8282225) to this batch as well.

Can do. I was actually wondering how much build noise all my pushes was creating, so that looks like it'll ease things a bit.

akashche:
Windows build setup with VS2017 [has worked for me](https://github.com/akashche/jdk8u-dev/actions/runs/1977187284): [modified submit.yml](https://github.com/akashche/jdk8u-dev/blob/gh-actions-win/.github/workflows/submit.yml), [additional file for a FreeType build](https://github.com/akashche/jdk8u-dev/blob/gh-actions-win/.github/workflows/freetype.vcxproj).

There is one questionable bit, when VS installer bootstrap launcher is being taken from [my msvs_2017_installer_bootstrap repo](https://github.com/akashche/msvs_2017_installer_bootstrap). This launcher is publicly available from MSFT ([download link](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=Community&rel=15)) but I was unable to automate its fetching easily and have uploaded it to this msvs_2017_installer_bootstrap personal repo.

jerboaa (approved):
This seems good enough for an initial integration. Better to have some (build) tests prior integration than nothing :)

One question: Do we really need file `.github/workflows/freetype.vcxproj`? If not, please remove.

gnu-andrew:
> This seems good enough for an initial integration. Better to have some (build) tests prior integration than nothing :)
> 
> One question: Do we really need file `.github/workflows/freetype.vcxproj`? If not, please remove.

It was provided by Alex for the Windows build: https://github.com/openjdk/jdk8u-dev/pull/3#issuecomment-1066169712

I'm no expert as to why this is needed. I did notice there were some FreeType build changes that could possibly be backported. I'd rather we did this in a follow-on PR though, so we can get the basics in. It's already been a month, and personally, I'm not going to have time to look at anything for the next few weeks, due to the security update.

akashche:
> > > This seems good enough for an initial integration. Better to have some (build) tests prior integration than nothing :)
> > > One question: Do we really need file `.github/workflows/freetype.vcxproj`? If not, please remove.
> > 
> > 
> > It was provided by Alex for the Windows build: [#3 (comment)](https://github.com/openjdk/jdk8u-dev/pull/3#issuecomment-1066169712)
> 
> OK. No expert on this either, so fine with me going in either way :-) It's certainly better than what we have currently.

This `freetype.vcxproj` file is used to replace the default FreeType build file. It has modifications to be buildable with VS2017. It is necessary for `--with-freetype-src` to work, with this flag FreeType is being built with msbuild (that used this build file) at configure time. We can avoid it by building FreeType by some other means (before running configure) and using `--with-freetype-lib` and `--with-freetype-include` instead. Just this `--with-freetype-src` seems to be "the most vanilla" way to build FreeType for 8u, so I suggest to continue using it.

gnu-andrew:
> > > This seems good enough for an initial integration. Better to have some (build) tests prior integration than nothing :)
> > > One question: Do we really need file `.github/workflows/freetype.vcxproj`? If not, please remove.
> > 
> > 
> > It was provided by Alex for the Windows build: [#3 (comment)](https://github.com/openjdk/jdk8u-dev/pull/3#issuecomment-1066169712)
> 
> OK. No expert on this either, so fine with me going in either way :-) It's certainly better than what we have currently.

Thanks. I've flagged the main bug for 8u approval now (I presume we can infer the others from that one).

gnu-andrew:
> > > > This seems good enough for an initial integration. Better to have some (build) tests prior integration than nothing :)
> > > > One question: Do we really need file `.github/workflows/freetype.vcxproj`? If not, please remove.
> > > 
> > > 
> > > It was provided by Alex for the Windows build: [#3 (comment)](https://github.com/openjdk/jdk8u-dev/pull/3#issuecomment-1066169712)
> > 
> > 
> > OK. No expert on this either, so fine with me going in either way :-) It's certainly better than what we have currently.
> 
> This `freetype.vcxproj` file is used to replace the default FreeType build file. It has modifications to be buildable with VS2017. It is necessary for `--with-freetype-src` to work, with this flag FreeType is being built with msbuild (that used this build file) at configure time. We can avoid it by building FreeType by some other means (before running configure) and using `--with-freetype-lib` and `--with-freetype-include` instead. Just this `--with-freetype-src` seems to be "the most vanilla" way to build FreeType for 8u, so I suggest to continue using it.

Thanks for the explanation. That does make things a lot clearer for me.

A couple of questions: 

1. Does this mean the file is not needed for the VS2010 build? I duplicated your changes in that build and it seems to work at present.
2. Any idea why 11u (which this is backported from) works without this? Does it build FreeType in a different way?

akashche:
> A couple of questions:
> 
> 1. Does this mean the file is not needed for the VS2010 build? I duplicated your changes in that build and it seems to work at present.

Yes, I believe it is not needed for VS2010. I haven't tried it with 32-bit VS2010 myself, but IIRC it worked out of the box with 64-bit VS2013 (freeware version of VS2010 doesn't support 64-bit, setup is possible with WinSDK7.1 toolchain, but it is complicated). Modified file was intended to change only 64-bit config (to make it working with 64-bit VS2017) so 32-bit config remained the same there. Thus `Copy FreeType project file` step on 32-bit most probably can be removed. 

> 
>  2. Any idea why 11u (which this is backported from) works without this? Does it build FreeType in a different way?

Yes, in 8u FreeType is not included in-tree, it was always used as an external dependency and for "open" jdk only (Oracle JDK didn't use it). `--with-freetype-src` flag is relatively recent, it was added in [JDK-8057538](https://bugs.openjdk.java.net/browse/JDK-8057538). In 11u FreeType is [included in-tree](https://github.com/openjdk/jdk11u/tree/master/src/java.desktop/share/native/libfreetype) and is built the same way as other in-tree dep libs.

gnu-andrew:
I see the bug now has `jdk8u-fix-yes`.

gnu-andrew:
> > A couple of questions:
> > 
> > 1. Does this mean the file is not needed for the VS2010 build? I duplicated your changes in that build and it seems to work at present.
> 
> Yes, I believe it is not needed for VS2010. I haven't tried it with 32-bit VS2010 myself, but IIRC it worked out of the box with 64-bit VS2013 (freeware version of VS2010 doesn't support 64-bit, setup is possible with WinSDK7.1 toolchain, but it is complicated). Modified file was intended to change only 64-bit config (to make it working with 64-bit VS2017) so 32-bit config remained the same there. Thus `Copy FreeType project file` step on 32-bit most probably can be removed.
> 

Thanks. We can maybe try that in a follow-up PR.

> > 2. Any idea why 11u (which this is backported from) works without this? Does it build FreeType in a different way?
> 
> Yes, in 8u FreeType is not included in-tree, it was always used as an external dependency and for "open" jdk only (Oracle JDK didn't use it). `--with-freetype-src` flag is relatively recent, it was added in [JDK-8057538](https://bugs.openjdk.java.net/browse/JDK-8057538). In 11u FreeType is [included in-tree](https://github.com/openjdk/jdk11u/tree/master/src/java.desktop/share/native/libfreetype) and is built the same way as other in-tree dep libs.

Interesting, thanks. It may be worth considering whether to bring the FreeType sources into 8u as well.

PR #25, opened 2022-03-29
8278138: OpenJDK8 fails to start on Windows 8.1 after upgrading compiler to VS2017

ktakakuri:
Could you please review the 8278138 bug fixes?

This problem is caused by missing DLL's which vcruntime140.dll depends on.
I think it is needed to copy the missing DLL's into jdk/bin as like jdk/jre/bin.
I fix to copy api-ms-win-*.dll files to jdk/bin when images directory is created.

### DIFF
--- a/jdk/make/Images.gmk
+++ b/jdk/make/Images.gmk
@@ -145,7 +145,8 @@ endif
 WINDOWS_JDK_BIN_FILES = \
     $(EXE_SUFFIX) \
     $(LIBRARY_PREFIX)jli$(SHARED_LIBRARY_SUFFIX) \
-    $(notdir $(MSVCR_DLL))
+    $(notdir $(MSVCR_DLL)) \
+    $(wildcard $(JDK_OUTPUTDIR)/bin/api-ms-win-*.dll)
 
 WINDOWS_JDKJRE_BIN_FILES := \
     $(LIBRARY_PREFIX)attach$(SHARED_LIBRARY_SUFFIX) \


ktakakuri:
Could anyone review the fix please?

ktakakuri:
Could someone please review this pull request?

jaokim (changes requested):
Hi,
Thank you for looking into this.
I've found that only copying the api-ms-* DLLs isn't sufficient. You also need to copy ucrtbase.dll among others. Instead of relying on the name, you can simply add all files from the the redistributable UCRT directory, in the same manner as in [CopyFiles.gmk.](https://github.com/openjdk/jdk8u-dev/blob/master/jdk/make/CopyFiles.gmk#L250)

ktakakuri:
> Hi, Thank you for looking into this. I've found that only copying the api-ms-* DLLs isn't sufficient. You also need to copy ucrtbase.dll among others. Instead of relying on the name, you can simply add all files from the the redistributable UCRT directory, in the same manner as in [CopyFiles.gmk.](https://github.com/openjdk/jdk8u-dev/blob/master/jdk/make/CopyFiles.gmk#L250)

Thank you for your review.
I think your comments are correct.

ktakakuri:
@jaokim 
I have committed your fix.
Could you please review it again?

jaokim (approved):
Changes looks good to me!

erikj79 (approved):

ktakakuri:
@erikj79 
Would you be willing to be a sponsor?

jaokim:
This needs to be approved according to https://wiki.openjdk.java.net/display/jdk8u/Main#Main-FixApprovals
I'm in the process of adding the needed labels to the JBS issue, and then a maintainer will take care of the approval.

ktakakuri:
@jaokim 
Thank you for adding the label.

@erikj79 
Would you please be a sponsor?

jerboaa:
@ktakakuri The issue this is referencing is resolved. I'm guessing it go resolved with a push to Oracle's JDK 8 fork. The commit for this isn't public so we need to do the SKARA trick to mention the backport bug like so: `Backport JDK-8278138`. It's strange that skara bots don't flag this with a warning (it's not labelled as a `backport`).

jerboaa:
Also, please enable PR checks. Perhaps a merge from master would do that.

jerboaa:
@ktakakuri Once this is being recognized as a backport I'll sponsor it for you.

PR #52, opened 2022-05-02
Backport 4d45c3ebc493bb2c85dab84b97840c8ba093ab1f

zhengyu123:
I would like to backport this patch to 8u for fixing a memory leak.

The original patch does not apply cleanly due to context difference. However, the patch is small, easily resolved manually.

Test:

- [x]  hotspot_compiler


phohensee (approved):
Lgtm.

zhengyu123:
Thanks for the review, @phohensee

gnu-andrew:
This has broken the build on AArch64: https://bugs.openjdk.java.net/browse/JDK-8287537

Not sure if it is related but this change uses `delete _overflow_arena;` while the original has `delete cb->_overflow_arena;`
I'm guessing the AArch64 build is trying to access something it does not expect to be NULL.

PR #55, opened 2022-05-09
Fix errors during builds without source tips

Roo4L:
If source tips files or directory is not provided, method GetSourceTips in
make/common/MakeBase.gmk would fail due to incorrect resolving of unquoted variables


dongbohe:
Hi, we also found this issue in our build environment, and have opened an issue [JDK-8286578](https://bugs.openjdk.java.net/browse/JDK-8286578) on JBS.
You can use this issue in your PR if you want:)

dongbohe:
You need to change the title of the PR to the same as the JBS, ie: `8286578: [8u] Build failure after rm -rf .git`

gnu-andrew:
#62 should fix this.

jerboaa:
Please test #62, which should fix your issues and would be the preferred approach as it also aligns this more how 11u works in this regard.

PR #64, opened 2022-05-23
8285727: [11u, 17u] Unify fix for JDK-8284920 with version

yan-too:
I need to backport this fix prepared by Christoph Langer to jdk8u, too. Without it, further backports using Token would be painful. The fix is not technically a backport from upstream for obvious reason so it would require review etc.

### DIFF
--- a/jaxp/src/com/sun/org/apache/xpath/internal/compiler/Lexer.java
+++ b/jaxp/src/com/sun/org/apache/xpath/internal/compiler/Lexer.java
@@ -358,7 +358,7 @@ class Lexer
 
         addToTokenQueue(pat.substring(i, i + 1));
         break;
-      case Token.COLON_CHAR:
+      case Token.COLON :
         if (i>0)
         {
           if (posOfNSSep == (i - 1))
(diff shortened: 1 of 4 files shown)


jerboaa:
@yan-too Could you please change the PR title to `Backport a95482acf83cc03bc562baace0d55d831d0b2b41` as it's really a backport of that change in JDK 17 (modulo path changes). Right now the bot complains about the issue not being open.

yan-too:
@jerboaa Actually that complain is quite common thing because  8285727 indeed is not open: it is Resolved/Fixed. I will change the title but I'm afraid there might be a problem in future since a95482acf83cc03bc562baace0d55d831d0b2b41 _is not in head_ as would be usual for the regular fixes (not backports) in 17. History stops in 17. Let's see though...

jerboaa:
> @jerboaa Actually that complain is quite common thing because 8285727 indeed is not open: it is Resolved/Fixed. I will change the title but I'm afraid there might be a problem in future since a95482acf83cc03bc562baace0d55d831d0b2b41 _is not in head_ as would be usual for the regular fixes (not backports) in 17. History of it stops in 17. Let's see though... it works! openjdk bot says "...with version from head" but has found the hash in 17u. Openjdk looks like hydra :-)

Thanks! Yeah, the original change was in 17 and downward, but it's still a backport of the JDK 17 unification change. jdk head got the same change with JDK-8284920.

jerboaa:
This is a clean backport modulo path changes. Trying to mark as such.

PR #67, opened 2022-05-30
Backport 22fad64529a890dd3ae8b07c7981d9a720cf8e96

yan-too:
I'd like to backport this jdk9 change to jdk8 because it would make backporting future important fixes and related tests easier. It is marked as Enhancement, however with time the absence of it would become nuisance. 
I've tested it so far with various XML-related tests. Tier tests are running.

### DIFF
--- a/jaxp/src/com/sun/org/apache/xpath/internal/compiler/FunctionTable.java
+++ b/jaxp/src/com/sun/org/apache/xpath/internal/compiler/FunctionTable.java
@@ -1,21 +1,23 @@
 /*
- * reserved comment block
- * DO NOT REMOVE OR ALTER!
+ * Copyright (c) 2015, Oracle and/or its affiliates. All rights reserved.
  */
-/*
- * Copyright 1999-2005 The Apache Software Foundation.
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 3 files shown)


gnu-andrew:
> Thank you @gnu-andrew ! As to the copyright, in the current (before the fix) version of jdk8u it is virtually the same as in 22fad6452 new part of the 1st Keywords.java hunk (and in 11u) barring some formatting difference (lines are shorter in 11u). I've used 22fad6452 so decided to skip the hunk altogether (the old part of the hunk differs significantly).

Right, what I'd suggest is just copying over Keywords.java from 11u. After this patch, they are the same bar minor formatting differences which came after this patch, which means `FUNC_HERE_STRING` is formatted differently to all the other constant declarations in the file (thank JDK-8068842 for this) I don't see any reason to keep these differences and they will only make any future patches which touch the file more difficult to apply.

> 
> The fixupFunctionTable() part, yes, I've missed it, thank you, retesting locally first...

Easy to miss as it was in a different commit. Version here looks fine with the logging changes.

yan-too:
You are right: this current state of copyright preamble etc. is a result of some reformatting in before-10 jdk, and nobody will backport that cleanup -- perhaps several of them -- as such. Updated.

PR #74, opened 2022-06-15
8288483: JAVA_TOOL_OPTIONS are silently truncated if they exceed 1024 bytes

dsamersoff:
The JAVA_TOOL_OPTIONS environment variable is used to pass additional JVM arguments.

The current implementation in jdk8 has an internal limit on the length of the option variable (1024 bytes) and the number of options (64).

A longer variable will be silently truncated and some options will be lost or the VM will exist with an unrecognized option error.

The fix is not a direct backport of the changes in the later JDKs, but I kept the code as close as possible to what we have in the latest jdk.

It's partial backport of relevant changes from JDK-8135195, JDK-8074895, JDK-8061999

---------
### Progress

### Issue
 * [JDK-8288483](https://bugs.openjdk.org/browse/JDK-8288483): JAVA_TOOL_OPTIONS are silently truncated if they exceed 1024 bytes

### Reviewers
 * [Paul Hohensee](https://openjdk.org/census#phh) (@phohensee - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/74/head:pull/74` \
`$ git checkout pull/74`

`$ git checkout pull/74` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/74/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 74`

`$ git pr show -t 74`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/74.diff">https://git.openjdk.org/jdk8u-dev/pull/74.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk8u-dev/pull/74#issuecomment-1156756160)

### DIFF
--- a/hotspot/src/share/vm/runtime/arguments.cpp
+++ b/hotspot/src/share/vm/runtime/arguments.cpp
@@ -3745,75 +3745,108 @@ jint Arguments::parse_java_tool_options_environment_variable(SysClassPath* scp_p
 }
 
 jint Arguments::parse_options_environment_variable(const char* name, SysClassPath* scp_p, bool* scp_assembly_required_p) {
-  const int N_MAX_OPTIONS = 64;
-  const int OPTION_BUFFER_SIZE = 1024;
-  char buffer[OPTION_BUFFER_SIZE];
+  char *buffer = ::getenv(name);
 
-  // The variable will be ignored if it exceeds the length of the buffer.
-  // Don't check this variable if user has special privileges
+  // Don't check this environment variable if user has special privileges
   // (e.g. unix su command).
-  if (os::getenv(name, buffer, sizeof(buffer)) &&
-      !os::have_special_privileges()) {
-    JavaVMOption options[N_MAX_OPTIONS];      // Construct option array
-    jio_fprintf(defaultStream::error_stream(),
-                "Picked up %s: %s\n", name, buffer);
-    char* rd = buffer;                        // pointer to the input string (rd)
-    int i;
-    for (i = 0; i < N_MAX_OPTIONS;) {         // repeat for all options in the input string
-      while (isspace(*rd)) rd++;              // skip whitespace
-      if (*rd == 0) break;                    // we re done when the input string is read completely
-
-      // The output, option string, overwrites the input string.
-      // Because of quoting, the pointer to the option string (wrt) may lag the pointer to
-      // input string (rd).
-      char* wrt = rd;
-
-      options[i++].optionString = wrt;        // Fill in option
-      while (*rd != 0 && !isspace(*rd)) {     // unquoted strings terminate with a space or NULL
-        if (*rd == '\'' || *rd == '"') {      // handle a quoted string
-          int quote = *rd;                    // matching quote to look for
-          rd++;                               // don't copy open quote
-          while (*rd != quote) {              // include everything (even spaces) up until quote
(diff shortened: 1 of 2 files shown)


phohensee (changes requested):
Nit: in arguments.hpp, move the declaration or parse_arguments_buffer down 2 lines after parse_java_options_environment variable to match tip's relative placement.

Otherwise fine.

phohensee (approved):
Lgtm.

jerboaa:
This seems a high-risk backport. As far as I can see it's not explained why it's needed. Especially why now? 8u has been around for some time without that fix. We cannot just backport everything.

gnu-andrew:
I tend to agree with Severin here.  Generally, enhancements shouldn't be backported.

Also, I don't see why this is being introduced under a new bug ID, and this is effectively hiding two enhancements inside a bug. The relevant fix for the 1024 limit is [JDK-8135195](https://bugs.openjdk.org/browse/JDK-8135195), but that issue is referring to a limit on an options file, which is not supported in 8u, as it doesn't have [JDK-8061999](https://bugs.openjdk.org/browse/JDK-8061999).  This patch doesn't backport the options file support, so is effectively treading new ground in lifting the limit just for the environment variable, but with most of the code changes from the aforementioned bugs.

I think they may be a case for a simpler fix that improves os::getenv so it warns when the environment variable passed is greater than the buffer length (probably only when Verbose is on). However, I don't see the case mentioned in the title of this PR; the variable is not truncated, it is simply ignored in its entirety if it is too long ("The variable will be ignored if it exceeds the length of the buffer.")

From `hotspot/src/os/linux/vm/os_linux.cpp`:

~~~
bool os::getenv(const char* name, char* buf, int len) {
  const char* val = ::getenv(name);
  if (val != NULL && strlen(val) < (size_t)len) {
    strcpy(buf, val);
    return true;
  }
  if (len > 0) buf[0] = 0;  // return a null string                                                                                                         
  return false;
}
~~~

So `os::getenv` will return `false`, just as it would if the environment variable was not defined.

dsamersoff:
The main rationally behind these changes is to fix spec violation. JDK 8 specification **does not set any limits for JAVA_TOOL_OPTIONS length and number of items**, so every time when a customer steps into this problem, it gets escalated, and it is not possible to argue that this is the expected OpenJDK behavior. 
This issue can happen, for example, when someone moves a setup with command line arguments pointing to a long path on the filesystem, or when moving an existing setup to buildpacks. From a regular user standpoint, the current behavior of simply skipping of some options leaves the user with unusable command line parameters. Worse, the user may not even notice that some options has been skipped. 

Please look at the testcase https://github.com/dsamersoff/jdk_8288483_scratchpad/pull/2/files (this testcase is not a part of PR because it doesn’t exist in the upstream)

Codewise, the code is essentially the same to what we have in 11u with some minimal changes required to accommodate API difference. 
The code in 11u is the result of several sequential changes that have been made under different CRs listed above. I chose to port the final code rather than come in with set of backports. But I can create a direct backport of code from jdk9 that was replaced later, if you think that it is less risky.

Please, look at https://github.com/dsamersoff/jdk_8288483_scratchpad/pull/1/files for difference between 11u (left) and what is proposed to 8u (right).

PR #88, opened 2022-07-21
8290832: It is no longer possible to change "user.dir" in the JDK8

mrserb:
This is the request to revert the fix for [JDK-8194154](https://bugs.openjdk.org/browse/JDK-8194154) integrated to the [openjdk8u342](https://github.com/openjdk/jdk8u/commit/5fcfb7acbe340c8347bd47413a36dd0b1e267dc1)

The fix for [JDK-8194154](https://bugs.openjdk.org/browse/JDK-8194154) disabled the possibility to change the "user.dir" property.
Changing the "user.dir" was not recommended from the beginning but it was not forbidden, so there are some old applications that rely on the old behavior. 

For compatibility reasons, it would be good to restore the old behavior.

Note that the test is simplified and saved under a different name, now it will check just the possibility to change the 'user.dir' using a simple path and name.

---------
### Progress

### Issue
 * [JDK-8290832](https://bugs.openjdk.org/browse/JDK-8290832): It is no longer possible to change "user.dir" in the JDK8

### Reviewers
 * [Christoph Langer](https://openjdk.org/census#clanger) (@RealCLanger - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev pull/88/head:pull/88` \
`$ git checkout pull/88`

`$ git checkout pull/88` \
`$ git pull https://git.openjdk.org/jdk8u-dev pull/88/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 88`

`$ git pr show -t 88`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/88.diff">https://git.openjdk.org/jdk8u-dev/pull/88.diff</a>

</details>


dsamersoff:
Looks good to me.

RealCLanger (commented):
The change itself looks good. But wouldn't it be better to do a clean backout of [JDK-8194154](https://bugs.openjdk.org/browse/JDK-8194154) and then add the test that user.dir can be changed in JDK8 with a different issue?

Or would it at least be possible that the test is not shown as renamed but rather as a deletion of the old test and the addition of a new test? (Not sure whether that's something that git did automatically or whether you can change this)

mrserb:
> Or would it at least be possible that the test is not shown as renamed but rather as a deletion of the old test and the addition of a new test? (Not sure whether that's something that git did automatically or whether you can change this)

It is actually a deletion in one commit and adding a new test by another, it is how the github shows it, but I can split it by different commits.

mrserb:
The patch is updated as requested, the test will be pushed separately.

RealCLanger (approved):
LGTM. You might consider renaming the JBS issue to something like "Backout JDK-8194154..." But I leave that to your or the 8u maintainers.

gnu-andrew:
Thanks for catching this @mrserb and the patch itself looks like a clean backout to me.

However, I suggest we do this on https://github.com/openjdk/jdk8u for an interim 8u345 release.  I've just requested that version from ops and will update the jcheck rules on 8u when that is created.

Presumably we'll do the test under a different bug?

mrserb:
> However, I suggest we do this on https://github.com/openjdk/jdk8u for an interim 8u345 release. I've just requested that version from ops and will update the jcheck rules on 8u when that is created.

Ok, I'll prepare the patch for that repo.

> Presumably we'll do the test under a different bug?

Yes, I'll add the test in a separate CR, so this patch will be as simple as "git revert"..

PR #106, opened 2022-08-19
8195607: sun/security/pkcs11/Secmod/TestNssDbSqlite.java failed with "NSS initialization failed" on NSS 3.34.1

martinuy:
Hi,

I'd like to propose the backport of the JDK-8195607 fix to 8u because this release is affected by this issue. Given that the legacy NSS DB has been deprecated for a long time and it's increasingly replaced by SQLite DBs, the chances that JDK-8 builds hit this problem are now higher.

The 11u backport applies cleanly, once paths are fixed.

To minimize risks, I manually debugged the OpenJDK code that connected to an NSS DB using the sql: prefix and everything was as expected.

I also run all the tests under the jdk/sun/security/pkcs11 category and noticed no-regressions strictly related to this fix. The test 'jdk/sun/security/pkcs11/Secmod/TrustAnchors.java' requires a special note, though. This test passes if run standalone but fails when run as part of a group. The reason is because when run as part of a group, other tests copy the file 'pkcs11.txt' to the temporary nssdb directory which is shared between all of them. This DB does not register the trust anchors module called 'Builtin Roots Module', which was available in the old secmod.db and is required by the test (which expects to see a Secmod$Module of TRUSTANCHOR type). While the test leaves the sun.security.pkcs11.SecmodTest::useSqlite field set to false and tries to access the legacy DB (also contained in the shared temporary nssdb directory), NSS realizes that there is a new SQLite DB in the same directory and reads the configuration from pkcs11.txt. Otherwise, if pkcs11.txt does not exist (as when the test is run standalone), it creates one with the information in secmod.db and it registers the 'Builtin Roots Module' module. All these findings have been verified with the 'modutil -list' command. While the issue is unrelated to what JDK-8195607 fixes, it has been uncovered by the existence of the pkcs11.txt file introduced with this change.

Why newer releases are not affected? Because jtreg runs tests (including TrustAnchors.java) in separated directories, so they don't share the nssdb temporary directory. In newer releases, TEST.ROOT has a requiredVersion attribute [1] indicating to jtreg that this can be done [2].

I've analyzed several ways to fix this issue but the one that I'd like to propose is to clean up the temporary nssdb directory for every test and avoid conflicts, giving every test the isolation that jtreg does when it sets test.classes uniquely per test in newer JDK releases.

Thanks,
Martin.-

--
[1] - https://github.com/openjdk/jdk11u/blob/master/test/jdk/TEST.ROOT#L61
[2] - https://github.com/openjdk/jtreg/blob/jtreg-6.2%2B1/src/share/classes/com/sun/javatest/regtest/config/Locations.java#L174

---------
### Progress

### Issue
 * [JDK-8195607](https://bugs.openjdk.org/browse/JDK-8195607): sun/security/pkcs11/Secmod/TestNssDbSqlite.java failed with "NSS initialization failed" on NSS 3.34.1

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev pull/106/head:pull/106` \
`$ git checkout pull/106`

`$ git checkout pull/106` \
`$ git pull https://git.openjdk.org/jdk8u-dev pull/106/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 106`

`$ git pr show -t 106`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/106.diff">https://git.openjdk.org/jdk8u-dev/pull/106.diff</a>

</details>

### DIFF
--- a/jdk/src/share/classes/sun/security/pkcs11/Secmod.java
+++ b/jdk/src/share/classes/sun/security/pkcs11/Secmod.java
@@ -206,7 +206,7 @@ public final class Secmod {
 
         if (configDir != null) {
             String configDirPath = null;
-            String sqlPrefix = "sql:/";
+            String sqlPrefix = "sql:";
             if (!configDir.startsWith(sqlPrefix)) {
                 configDirPath = configDir;
             } else {
--- a/jdk/src/share/native/sun/security/pkcs11/j2secmod.c
+++ b/jdk/src/share/native/sun/security/pkcs11/j2secmod.c
@@ -69,9 +69,14 @@ JNIEXPORT jboolean JNICALL Java_sun_security_pkcs11_Secmod_nssInitialize
     int res = 0;
     FPTR_Initialize initialize =
         (FPTR_Initialize)findFunction(env, jHandle, "NSS_Initialize");
+    #ifdef SECMOD_DEBUG
+    FPTR_GetError getError =
+        (FPTR_GetError)findFunction(env, jHandle, "PORT_GetError");
+    #endif // SECMOD_DEBUG
     unsigned int flags = 0x00;
     const char *configDir = NULL;
     const char *functionName = NULL;
+    const char *configFile = NULL;
 
     /* If we cannot initialize, exit now */
     if (initialize == NULL) {
@@ -97,13 +102,18 @@ JNIEXPORT jboolean JNICALL Java_sun_security_pkcs11_Secmod_nssInitialize
         flags = 0x20; // NSS_INIT_OPTIMIZESPACE flag
     }
 
+    configFile = "secmod.db";
+    if (configDir != NULL && strncmp("sql:", configDir, 4U) == 0) {
+        configFile = "pkcs11.txt";
+    }
+
     /*
      * If the NSS_Init function is requested then call NSS_Initialize to
      * open the Cert, Key and Security Module databases, read only.
      */
     if (strcmp("NSS_Init", functionName) == 0) {
         flags = flags | 0x01; // NSS_INIT_READONLY flag
-        res = initialize(configDir, "", "", "secmod.db", flags);
+        res = initialize(configDir, "", "", configFile, flags);
 
     /*
      * If the NSS_InitReadWrite function is requested then call
@@ -111,7 +121,7 @@ JNIEXPORT jboolean JNICALL Java_sun_security_pkcs11_Secmod_nssInitialize
      * read/write.
      */
     } else if (strcmp("NSS_InitReadWrite", functionName) == 0) {
-        res = initialize(configDir, "", "", "secmod.db", flags);
+        res = initialize(configDir, "", "", configFile, flags);
 
     /*
      * If the NSS_NoDB_Init function is requested then call
@@ -137,6 +147,13 @@ cleanup:
         (*env)->ReleaseStringUTFChars(env, jConfigDir, configDir);
     }
     dprintf1("-res: %d\n", res);
+    #ifdef SECMOD_DEBUG
+    if (res == -1) {
+        if (getError != NULL) {
+            dprintf1("-NSS error: %d\n", getError());
+        }
+    }
+    #endif // SECMOD_DEBUG
 
     return (res == 0) ? JNI_TRUE : JNI_FALSE;
 }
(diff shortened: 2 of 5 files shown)


jerboaa:
> The 11u backport applies cleanly, once paths are fixed.

Yes, that seems to be the case if https://github.com/openjdk/jdk8u-dev/pull/106/commits/73173863ab9da7ee2ec080401180475bf7ee22e1 is to be looked at on its own.
 
> To minimize risks, I manually debugged the OpenJDK code that connected to an NSS DB using the sql: prefix and everything was as expected.

OK.

> I also run all the tests under the jdk/sun/security/pkcs11 category and noticed no-regressions strictly related to this fix. The test 'jdk/sun/security/pkcs11/Secmod/TrustAnchors.java' requires a special note, though. This test passes if run standalone but fails when run as part of a group. The reason is because when run as part of a group, other tests copy the file 'pkcs11.txt' to the temporary nssdb directory which is shared between all of them. This DB does not register the trust anchors module called 'Builtin Roots Module', which was available in the old secmod.db and is required by the test (which expects to see a Secmod$Module of TRUSTANCHOR type). While the test leaves the sun.security.pkcs11.SecmodTest::useSqlite field set to false and tries to access the legacy DB (also contained in the shared temporary nssdb directory), NSS realizes that there is a new SQLite DB in the same directory and reads the configuration from pkcs11.txt. Otherwise, if pkcs11.txt does not exist (as when the test is run standalone), it creates one with the information in secmod.db and it registers the 'Builtin Roots Module' module. All these findings have been verified with the 'modutil -list' command. While the issue is unrelated to what JDK-8195607 fixes, it has been uncovered by the existence of the pkcs11.txt file introduced with this change.
> 
> Why newer releases are not affected? Because jtreg runs tests (including TrustAnchors.java) in separated directories, so they don't share the nssdb temporary directory. In newer releases, TEST.ROOT has a requiredVersion attribute [1] indicating to jtreg that this can be done [2].
> 
> I've analyzed several ways to fix this issue but the one that I'd like to propose is to clean up the temporary nssdb directory for every test and avoid conflicts, giving every test the isolation that jtreg does when it sets test.classes uniquely per test in newer JDK releases.

This seems strange, wouldn't a jtreg version `> 4.2 b08` not run into this problem? I.e. if I run with jtreg 5 (oldest version I have here), I wouldn't see this issue? If so I'm not sure this is worth adding https://github.com/openjdk/jdk8u-dev/pull/106/commits/22c1df05d2ab51a400ebdc69566e84fdaf32efec to this backport. It's not in head either. So at the very least this should be an 8u-only bug. But if that's only happening with jtreg 4, I'd say this is not worth fixing at all (or perhaps add jtreg 5 as a minimal version. Everyone should be on newer versions these days anyway.

martinuy:
Hi @jerboaa ,

Thanks for your review.

Let me clarify the jtreg issue. It's not that the test fails because it's run with an old jtreg. In fact, I'm running with jtreg 5 and it's failing. The problem is that the TEST.ROOT in JDK-8 does not set a "minimum jtreg required" (probably because it's a newer feature than JDK-8). In these cases, jtreg takes a conservative approach and does not split the text execution into different directories. Setting a "minimum jtreg required" in JDK-8 would be a high risk as we don't know the impact of other tests in newer or older jtreg releases. Having this test broken would be a regression. I hope we can have this fix, or you might come up with a different one. Let me know if that helps for a better understanding.

Martin.-

jerboaa:
> Let me clarify the jtreg issue. It's not that the test fails because it's run with an old jtreg. In fact, I'm running with jtreg 5 and it's failing. The problem is that the TEST.ROOT in JDK-8 does not set a "minimum jtreg required" (probably because it's a newer feature than JDK-8). In these cases, jtreg takes a conservative approach and does not split the text execution into different directories. Setting a "minimum jtreg required" in JDK-8 would be a high risk as we don't know the impact of other tests in newer or older jtreg releases. Having this test broken would be a regression. I hope we can have this fix, or you might come up with a different one. Let me know if that helps for a better understanding.

Thanks for the clarification. I'm fine with your approach, but https://github.com/openjdk/jdk8u-dev/commit/22c1df05d2ab51a400ebdc69566e84fdaf32efec should go in as a separate 8u-specific patch. Or we could take this as an opportunity to require jtreg 5+ (which should be fine IMHO, I don't agree it's a high risk change). Either way this should be a separate discussion with a separate bug.

martinuy:
Hi @jerboaa ,

Thanks for your comment.

I've committed a revert for https://github.com/openjdk/jdk8u-dev/commit/22c1df05d2ab51a400ebdc69566e84fdaf32efec and opened a new ticket for it: https://bugs.openjdk.org/browse/JDK-8292998

I'll not propose this backport to 8u until I have JDK-8292998 integrated.

jerboaa:
Looks like a clean backport to me now.

gnu-andrew (approved):
Backport looks clean.
8u-dev is currently closed for rampdown. Please wait for it to reopen before integrating.

PR #111, opened 2022-08-23
Backport 844c504bc18fa8a79ceacc6b6f235650f5e643bf

apavlyutkin:
Hi!

Here is a backport of JDK-8271459 that fixes NegativeArraySizeException on using StringBuilder of negative capacity. The patch from jdk-11 is applied cleanly except the path suffling.

Verification (amd64/Ubuntu 20.04): the reproducer from JBS caught 10000 of 10000 exceptions
Regression: runtime


apavlyutkin:
ping

jerboaa:
@rwestrel Could you please review this? Does that look OK to you for 8u inclusion?

jerboaa:
> @jerboaa backporting this is reasonable I think.

Thanks!

apavlyutkin:
> That looks good to me. Why not include the test case?

Original tests were not backported to the 11 because they are based on IR test framework that does not exists in the 11. I raised JDK-8271459 to re-factor the tests and incorporate them into the 11 and the 8.

rwestrel:
> > That looks good to me. Why not include the test case?
> 
> Original tests were not backported to the 11 because they are based on IR test framework that does not exists in the 11. I raised JDK-8295322 to re-factor the tests and incorporate them into the 11 and the 8.

Makes sense.

rwestrel (approved):

PR #115, opened 2022-08-26
JDK-8292998: Clean Secmod temporary NSS DB directory before test execution

martinuy:
Hello,

I'd like to propose this fix as described in https://bugs.openjdk.org/browse/JDK-8292998

The temporary directory used by Secmod tests to copy NSS DB files is deleted (which includes all its files and, potentially, subdirectories) and created empty again before running the test. I decided to implement the recursive delete function here for 3 reasons: 1) it's a short and simple function,  2) convenience (we don't have to link every current Secmod test to a test library) and 3) increase chances of backporting future Secmod tests cleanly. Contrary to the approach of setting a minimum jtreg version required, the scope of this change are Secmod tests only and the risk should be minimal.

Thanks,
Martin.-

---------
### Progress
- [ ] Change must be properly reviewed (2 reviews required, with at least 2 [Reviewers](https://openjdk.org/bylaws#reviewer))

### Issue
 * [JDK-8292998](https://bugs.openjdk.org/browse/JDK-8292998): Clean Secmod temporary NSS DB directory before test execution

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev pull/115/head:pull/115` \
`$ git checkout pull/115`

`$ git checkout pull/115` \
`$ git pull https://git.openjdk.org/jdk8u-dev pull/115/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 115`

`$ git pr show -t 115`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/115.diff">https://git.openjdk.org/jdk8u-dev/pull/115.diff</a>

</details>

### DIFF
--- a/jdk/test/sun/security/pkcs11/SecmodTest.java
+++ b/jdk/test/sun/security/pkcs11/SecmodTest.java
@@ -26,6 +26,11 @@
 
 import java.io.*;
 
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
 import java.security.Provider;
 
 public class SecmodTest extends PKCS11Test {
@@ -60,9 +65,10 @@ public class SecmodTest extends PKCS11Test {
             System.setProperty("pkcs11test.nss.db", DBDIR);
         }
         File dbdirFile = new File(DBDIR);
-        if (dbdirFile.exists() == false) {
-            dbdirFile.mkdir();
+        if (dbdirFile.exists()) {
+            deleteDir(dbdirFile.toPath());
         }
+        dbdirFile.mkdir();
 
         if (useSqlite) {
             copyFile("key4.db", BASE, DBDIR);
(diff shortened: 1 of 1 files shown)


jerboaa (approved):
This looks OK to me. Given that this change makes test runs not leak files from one test to another this is good! Later JDK versions are doing the same, but with a different means. Thumbs up.

zzambers:
I wander why "DBDIR" is created inside of "classes.dir". Apart from not being cleaned up this also seems like potential source of problems when tests are run concurrently. Couldn't "DBDIR" just be placed in "scratch" [1] directory? Or is there something I am missing here?

[1] https://openjdk.org/jtreg/faq.html#scratch-directory

martinuy:
Hi @zzambers ,

It's a good point. I think that Secmod tests are not and have never been ready to run in parallel, at least not in 8u. As an example, they copy entire files to the same location without any file system synchronization scheme, which can potentially lead to inconsistencies. What we have fixed here is the sequential run of these tests, which was working before JDK-8195607.

jdk/test/sun/security/pkcs11/SecmodTest.java
```diff
@@ -90,6 +96,25 @@ private static void copyFile(String name, String srcDir, String dstDir) throws I
         out.close();
     }
 
+    private static void deleteDir(final Path directory) throws IOException {
```
gnu-andrew, line 99: Can we not use `FileUtils.deleteFileTreeWithRetry(Path)` here? That seems to be the same thing, but with better error handling.

See `jdk/test/lib/jdk/test/lib/util/FileUtils.java`

gnu-andrew (changes requested):

zzambers:
@martinuy I was just thinking if just turning this line
`DBDIR = System.getProperty("test.classes", ".") + SEP + "tmpdb";`
into
`DBDIR = "." + SEP + "tmpdb";`
instead of doing manual cleanup wouldn't be simpler and more robust solution (using "scratch" dir). But I have not actually tested that...

PR #125, opened 2022-09-29
8253702: BigSur version number reported as 10.16, should be 11.nn

omikhaltsova:
I'd like to backport JDK-8253702 to jdk8u in order to fix the incorrect version detection for macOS.

The patch applied manually due to the location difference:
`src/java.base/macosx/native/libjava/java_props_macosx.c` (source)
`jdk/src/solaris/native/java/lang/java_props_macosx.c` (destination)
But it's small and low risk. The changes are related only to one function `void setOSNameAndVersion(..)` within the single file. The patch is identical to the original one.

Tested with a simple test printing the OS version on macOS Big Sur:
```
public class Main {
        System.out.println(System.getProperty("os.version"));
    }
}
```

---------
### Progress

### Issue
 * [JDK-8253702](https://bugs.openjdk.org/browse/JDK-8253702): BigSur version number reported as 10.16, should be 11.nn

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev pull/125/head:pull/125` \
`$ git checkout pull/125`

`$ git checkout pull/125` \
`$ git pull https://git.openjdk.org/jdk8u-dev pull/125/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 125`

`$ git pr show -t 125`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/125.diff">https://git.openjdk.org/jdk8u-dev/pull/125.diff</a>

</details>

### DIFF
--- a/jdk/src/solaris/native/java/lang/java_props_macosx.c
+++ b/jdk/src/solaris/native/java/lang/java_props_macosx.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2017, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


jerboaa:
Why is this handling two backports with a single patch? You can do dependent prs if you wish, but please keep the patches separate. Thanks!

omikhaltsova:
Thanks for taking a look! It seemed to me a good idea to backport these 2 small patches together because they fix the similar bugs, locate in the same place and it'll be easy to review them also together. Anyway I will separate them if it's better.

jerboaa:
Path changes only, marking as clean.

PR #127, opened 2022-10-03
8230305: Cgroups v2: Container awareness

jmtd:
This is a backport of 8230305 (Cgroups v2: Container awareness) for JDK8u, working from the jdk11u-dev backport as part of an effort to backport cgroups v2 to jdk8u-dev.

The patch does not apply clean after unshuffling and path fixing:

 * mostly copyright line issues
 * different package name for jdk.test.lib.process.OutputAnalyzer
 * Most of hotspot/src/os/linux/vm/cgroupSubsystem_linux.cpp failed due to context changes, manually resolved (mostly deletions)

A few other changes were made

 *  Rework use of newer logging API (log_debug etc) to use guarded tty->print_cr
 *  replace os::strerror with the libc strerror (which is thread unsafe, but that should be addressed by backporting 8148425 separately)

patch touched hotspot/test/runtime/containers/docker/TestCPUAwareness.java which passes afterwards; I'm running further tests now (and awaiting GitHub CI doing the same)

---------
### Progress

### Issues
 * [JDK-8230305](https://bugs.openjdk.org/browse/JDK-8230305): Cgroups v2: Container awareness
 * [JDK-8216366](https://bugs.openjdk.org/browse/JDK-8216366): Add rationale to PER_CPU_SHARES define
 * [JDK-8229202](https://bugs.openjdk.org/browse/JDK-8229202): Docker reporting causes secondary crashes in error handling
 * [JDK-8232207](https://bugs.openjdk.org/browse/JDK-8232207): Linux os::available_memory re-reads cgroup configuration on every invocation
 * [JDK-8254997](https://bugs.openjdk.org/browse/JDK-8254997): Remove unimplemented OSContainer::read_memory_limit_in_bytes

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev pull/127/head:pull/127` \
`$ git checkout pull/127`

`$ git checkout pull/127` \
`$ git pull https://git.openjdk.org/jdk8u-dev pull/127/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 127`

`$ git pr show -t 127`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/127.diff">https://git.openjdk.org/jdk8u-dev/pull/127.diff</a>

</details>

### DIFF
--- a/hotspot/src/os/linux/vm/cgroupSubsystem_linux.cpp
+++ b/hotspot/src/os/linux/vm/cgroupSubsystem_linux.cpp
@@ -28,7 +28,6 @@
 #include "cgroupSubsystem_linux.hpp"
 #include "cgroupV1Subsystem_linux.hpp"
 #include "cgroupV2Subsystem_linux.hpp"
-#include "logging/log.hpp"
 #include "memory/allocation.hpp"
 #include "runtime/globals.hpp"
 #include "runtime/os.hpp"
@@ -65,8 +64,9 @@ CgroupSubsystem* CgroupSubsystemFactory::create() {
    */
   cgroups = fopen("/proc/cgroups", "r");
   if (cgroups == NULL) {
-      log_debug(os, container)("Can't open /proc/cgroups, %s",
-                               os::strerror(errno));
+      if(PrintContainerInfo)
+        tty->print_cr("Can't open /proc/cgroups, %s",
+                                 os::strerror(errno));
       return NULL;
   }
 
@@ -108,7 +108,8 @@ CgroupSubsystem* CgroupSubsystemFactory::create() {
 
   if (!all_controllers_enabled) {
     // one or more controllers disabled, disable container support
-    log_debug(os, container)("One or more required controllers disabled at kernel level.");
+    if(PrintContainerInfo)
+      tty->print_cr("One or more required controllers disabled at kernel level.");
     return NULL;
   }
 
@@ -120,8 +121,9 @@ CgroupSubsystem* CgroupSubsystemFactory::create() {
    */
   cgroup = fopen("/proc/self/cgroup", "r");
   if (cgroup == NULL) {
-    log_debug(os, container)("Can't open /proc/self/cgroup, %s",
-                             os::strerror(errno));
+    if(PrintContainerInfo)
+      tty->print_cr("Can't open /proc/self/cgroup, %s",
+                               os::strerror(errno));
     return NULL;
   }
 
@@ -169,8 +171,9 @@ CgroupSubsystem* CgroupSubsystemFactory::create() {
     // Find the cgroup2 mount point by reading /proc/self/mountinfo
     mntinfo = fopen("/proc/self/mountinfo", "r");
     if (mntinfo == NULL) {
-        log_debug(os, container)("Can't open /proc/self/mountinfo, %s",
-                                 os::strerror(errno));
+        if(PrintContainerInfo)
+          tty->print_cr("Can't open /proc/self/mountinfo, %s",
+                                   os::strerror(errno));
         return NULL;
     }
 
@@ -192,7 +195,8 @@ CgroupSubsystem* CgroupSubsystemFactory::create() {
     }
     fclose(mntinfo);
     if (!mount_point_found) {
-      log_trace(os, container)("Mount point for cgroupv2 not found in /proc/self/mountinfo");
+      if(PrintContainerInfo)
+        tty->print_cr("Mount point for cgroupv2 not found in /proc/self/mountinfo");
       return NULL;
     }
     // Cgroups v2 case, we have all the info we need.
@@ -204,12 +208,14 @@ CgroupSubsystem* CgroupSubsystemFactory::create() {
       os::free(cg_infos[i]._name);
       os::free(cg_infos[i]._cgroup_path);
     }
-    log_debug(os, container)("Detected cgroups v2 unified hierarchy");
+    if(PrintContainerInfo)
+      tty->print_cr("Detected cgroups v2 unified hierarchy");
     return new CgroupV2Subsystem(unified);
   }
 
   // What follows is cgroups v1
-  log_debug(os, container)("Detected cgroups hybrid or legacy hierarchy, using cgroups v1 controllers");
+  if(PrintContainerInfo)
+    tty->print_cr("Detected cgroups hybrid or legacy hierarchy, using cgroups v1 controllers");
 
   /*
    * Find the cgroup mount point for memory and cpuset
@@ -223,8 +229,9 @@ CgroupSubsystem* CgroupSubsystemFactory::create() {
    */
   mntinfo = fopen("/proc/self/mountinfo", "r");
   if (mntinfo == NULL) {
-      log_debug(os, container)("Can't open /proc/self/mountinfo, %s",
-                               os::strerror(errno));
+      if(PrintContainerInfo)
+        tty->print_cr("Can't open /proc/self/mountinfo, %s",
+                                 os::strerror(errno));
       return NULL;
   }
 
@@ -253,19 +260,23 @@ CgroupSubsystem* CgroupSubsystemFactory::create() {
   fclose(mntinfo);
 
   if (memory == NULL) {
-    log_debug(os, container)("Required cgroup v1 memory subsystem not found");
+    if(PrintContainerInfo)
+      tty->print_cr("Required cgroup v1 memory subsystem not found");
     return NULL;
   }
   if (cpuset == NULL) {
-    log_debug(os, container)("Required cgroup v1 cpuset subsystem not found");
+    if(PrintContainerInfo)
+      tty->print_cr("Required cgroup v1 cpuset subsystem not found");
     return NULL;
   }
   if (cpu == NULL) {
-    log_debug(os, container)("Required cgroup v1 cpu subsystem not found");
+    if(PrintContainerInfo)
+      tty->print_cr("Required cgroup v1 cpu subsystem not found");
     return NULL;
   }
   if (cpuacct == NULL) {
-    log_debug(os, container)("Required cgroup v1 cpuacct subsystem not found");
+    if(PrintContainerInfo)
+      tty->print_cr("Required cgroup v1 cpuacct subsystem not found");
     return NULL;
   }
 
@@ -356,7 +367,8 @@ int CgroupSubsystem::active_processor_count() {
   CachedMetric* cpu_limit = contrl->metrics_cache();
   if (!cpu_limit->should_check_metric()) {
     int val = (int)cpu_limit->value();
-    log_trace(os, container)("CgroupSubsystem::active_processor_count (cached): %d", val);
+    if(PrintContainerInfo)
+      tty->print_cr("CgroupSubsystem::active_processor_count (cached): %d", val);
     return val;
   }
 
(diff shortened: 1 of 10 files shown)


jmtd:
Local test failure observed in runtime/containers/docker/TestMemoryAwareness.java:
```
JavaTest Message: Test threw exception: java.lang.RuntimeException: 'Memory Soft Limit.*524288000' missing from stdout/stderr 
```
this might be [JDK-8253714](https://bugs.openjdk.org/browse/JDK-8253714), which is on the list to backport, I'll look at this next.

hotspot/src/os/linux/vm/cgroupV1Subsystem_linux.hpp
```diff
@@ -0,0 +1,118 @@
...
+    CachingCgroupController * cpu_controller() { return _cpu; }
+
+  private:
+    julong _unlimited_memory;
+
+    /* controllers */
+    CachingCgroupController* _memory = NULL;
+    CgroupV1Controller* _cpuset = NULL;
+    CachingCgroupController* _cpu = NULL;
+    CgroupV1Controller* _cpuacct = NULL;
```
poonamparhar, line 103: For arm32 platform, if the c++ compiler version being used is earlier than 11, then the non-static member initialization will cause compilation failure.

jmtd: Thanks for pointing this out! I'll investigate how to best address that in this patch series.

jerboaa:
@jmtd Any particular reason this PR depends on the Metrics PR? In JDK 11u and JDK head this was the first one of the cg v2 patch series. If you agree, please drop the dependency, targeting `master` directly.

hotspot/src/os/linux/vm/cgroupSubsystem_linux.hpp
```diff
@@ -0,0 +1,273 @@
...
+ *
+ * [1] https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-cpu
+ *     In particular:
+ *        When using Docker:
+ *          The spec.containers[].resources.requests.cpu is converted to its core value, which is potentially
+ *          fractional, and multiplied by 1024. The greater of this number or 2 is used as the value of the
+ *          --cpu-shares flag in the docker run command.
+ * [2] https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerDefinition.html
+ * [3] https://github.com/apache/mesos/blob/3478e344fb77d931f6122980c6e94cd3913c441d/src/docker/docker.cpp#L648
+ *     https://github.com/apache/mesos/blob/3478e344fb77d931f6122980c6e94cd3913c441d/src/slave/containerizer/mesos/isolators/cgroups/constants.hpp#L30
```
jerboaa, line 52: This is part of `JDK-8216366` which isn't yet in 8u. Please add the issue to the commit message by doing:

```
/issue add JDK-8216366
```

hotspot/src/os/linux/vm/osContainer_linux.hpp
```diff
@@ -66,8 +65,7 @@ class OSContainer: AllStatic {
 };
 
 inline bool OSContainer::is_containerized() {
-  assert(_is_initialized, "OSContainer not initialized");
```
jerboaa, line 69: This hunk got removed with `JDK-8229202` which is not yet part of 8u. Please add this issue by:

```
/issue add JDK-8229202
```

hotspot/src/os/linux/vm/cgroupSubsystem_linux.cpp
```diff
@@ -0,0 +1,436 @@
...
+ *    memory limit in bytes or
+ *    -1 for unlimited
+ *    OSCONTAINER_ERROR for not supported
+ */
+jlong CgroupSubsystem::memory_limit_in_bytes() {
+  CachingCgroupController* contrl = memory_controller();
+  CachedMetric* memory_limit = contrl->metrics_cache();
+  if (!memory_limit->should_check_metric()) {
+    return memory_limit->value();
+  }
```
jerboaa, line 431: This is actually code added with [JDK-8232207](https://bugs.openjdk.org/browse/JDK-8232207) which shouldn't affect JDK 8. So I'm not sure about this. Add it to 8u anyway (and have less divergence) or remove and deal with the code difference.

Whatever the way, we need to make this clear by: a) adding the issue using `/issue add` or b) restructuring this code to account for it. I'm slightly in favor of b).

hotspot/src/os/linux/vm/cgroupSubsystem_linux.cpp
```diff
@@ -0,0 +1,436 @@
...
+int CgroupSubsystem::active_processor_count() {
+  int quota_count = 0, share_count = 0;
+  int cpu_count, limit_count;
+  int result;
+
+  // We use a cache with a timeout to avoid performing expensive
+  // computations in the event this function is called frequently.
+  // [See 8227006].
+  CachingCgroupController* contrl = cpu_controller();
+  CachedMetric* cpu_limit = contrl->metrics_cache();
```
jerboaa, line 367: For the processor counts we do want the caching so as to not regress JDK-8227006. For memory on the other hand we should probably remove it in the 8u backport.

jerboaa (changes requested):
This seems promising, thanks. Please use explicit parenthesis for `PrintContainerInfo` log lines throughout. Also with a space before and after the `if`. We need to account for https://bugs.openjdk.org/browse/JDK-8227006 (which is in 8u) and https://bugs.openjdk.org/browse/JDK-8232207 (not currently in 8u), though. This patch would include both.

jmtd:
> @jmtd Any particular reason this PR depends on the Metrics PR? In JDK 11u and JDK head this was the first one of the cg v2 patch series. If you agree, please drop the dependency, targeting master directly.

I collapsed the patch dependency tree into a linear list, and as a consequence some patches have GitHub PR dependencies which are artificial. This one (depending on  8231111 (pr/121)) is an example; it's because the downstream bug 8253714 (pr/130) depends on both.

I could flip the artifical dependency around, and make 8231111 (metrics, pr/121) depend on this (8230305, pr/127) instead (and follow that change down the chain). Would you prefer that?

jerboaa:
> I could flip the artifical dependency around, and make 8231111 (metrics, pr/121) depend on this (8230305, pr/127) instead (and follow that change down the chain). Would you prefer that?

Yes, please. That's matching how patches got integrated in JDK head and JDK 11 updates.

jmtd:
> Please use explicit parenthesis for PrintContainerInfo log lines throughout. Also with a space before and after the if

Sure! I'm assuming you mean spaces and braces like this

```
@@ -64,9 +64,10 @@ CgroupSubsystem* CgroupSubsystemFactory::create() {
    */
   cgroups = fopen("/proc/cgroups", "r");
   if (cgroups == NULL) {
-      if(PrintContainerInfo)
+      if (PrintContainerInfo) {
         tty->print_cr("Can't open /proc/cgroups, %s",
                                  strerror(errno));
+      }
       return NULL;
   }
```

(although I'm not sure what you mean by spaces before the `if`) and I'm working that through now.

hotspot/src/os/linux/vm/cgroupV1Subsystem_linux.hpp
jmtd: I've spent some time trying to understand this issue, and this is my current understanding of it. ARM32 is the circumstance which surfaced the problem, but this is not an ARM32 problem. The syntax here was introduced in C++11. The minimal (GCC) compiler version stated for JDK8u is GCC 4.3, which has only partial, experimental support for C++11. You didn't state what compiler you were using but I'm going to assume it's one in the supported list for jdk8u. What I take from this is, JDK8u can't use all C++11 features, since some of the supported compiler-versions don't have all C++11 features; and in particular this feature.

These lines of code are unchanged in jdk mainline at the moment, however, jdk has moved forward the minimal supported compiler version. I doubt that a fix for this would be received favourably in jdk mainline, so we probably need to address it in jdk8u-dev directly (not via backport).

poonamparhar: The build where I saw compilation failures was on a 32-bit platform with C++ compiler version 4.7.2. Here's the error message: 
error: non-static data member initializers only available with -std=c++11 or -std=gnu++11 [-Werror]

In our jdk8 backport, we initialize these data members using constructor initializer-list.

jmtd: I've just pushed a commit that should resolve this (e6fcc66863e). Thanks for bringing it up!

hotspot/src/os/linux/vm/cgroupSubsystem_linux.cpp
jmtd: I tried a very basic JMH bench with/without the memory caching in place and it had no obvious impact one way or another, but I'm leaning towards a) because removing it means re-working quite a few bits, including switching _memory from a CachingCgroupController to CgroupController (or CgroupV1Controller etc) so the macros GET_CONTAINER_INFO and friends still work. The deltas will be across at least four source files.

jmtd:
/issue add JDK-8232207

@jerboaa , if you disagree with adding this issue and want it backed out afterall, please let me know!

jmtd:
As discussed, and now that all open review notes were addressed, I've re-parented this PR onto master.

jmtd:
After reparenting this PR to master last night, I've inadvertently pulled in the commits from the previous base (PR for 8231111). I'm working on filtering them out now.

hotspot/src/os/linux/vm/cgroupSubsystem_linux.cpp
jerboaa: Thanks. My thinking was for b) to use a patch like:

```
commit 9546da23b777625cb2bdd397080dae280e91b854 (HEAD -> pull/127-rebased)
Author: Severin Gehwolf <sgehwolf@redhat.com>
Date:   Thu Nov 3 19:55:20 2022 +0100

    Only cache CPU metrics

diff --git a/hotspot/src/os/linux/vm/cgroupSubsystem_linux.cpp b/hotspot/src/os/linux/vm/cgroupSubsystem_linux.cpp
index b93c8b308b..923ad7838a 100644
--- a/hotspot/src/os/linux/vm/cgroupSubsystem_linux.cpp
+++ b/hotspot/src/os/linux/vm/cgroupSubsystem_linux.cpp
@@ -440,13 +440,6 @@ int CgroupSubsystem::active_processor_count() {
  *    OSCONTAINER_ERROR for not supported
  */
 jlong CgroupSubsystem::memory_limit_in_bytes() {
-  CachingCgroupController* contrl = memory_controller();
-  CachedMetric* memory_limit = contrl->metrics_cache();
-  if (!memory_limit->should_check_metric()) {
-    return memory_limit->value();
-  }
   jlong mem_limit = read_memory_limit_in_bytes();
-  // Update cached metric to avoid re-reading container settings too often
-  memory_limit->set_value(mem_limit, OSCONTAINER_CACHE_TIMEOUT);
   return mem_limit;
 }
```

But my basic experiments don't suggest this makes a whole lot of difference. Given that `available_memory()` is being used in `hotspot/src/share/vm/jfr/periodic/jfrPeriodic.cpp` for the physical memory event, I'm OK with leaving it in as it might be beneficial for the JFR tracing. If that turns out to be a problem, we can revisit this later. OK with keeping it as is.

jerboaa (approved):
This seems fine now.

jmtd:
/issue add 8254997

This PR has most of [8232207](https://bugs.openjdk.org/browse/JDK-8232207) in it, so we added that issue. 8232207 originally added a line in osContainer_linux.hpp which was later removed in 8254997. This PR does not add the line, so we address 8254997 at the same time.

gnu-andrew (approved):
Backport looks good. It's not clear to me why JDK-8254997 needed to be bundled into this change rather than a follow-up, but it's not worth changing now.

Approved for 8u.

PR #131, opened 2022-10-05
8294863: Enable partial tier1 testing in GHA for JDK8

zzambers:
This change adds support for tier1 testing in github actions by fixing workflow to work  with JDK8. (Changes were needed due to build system difference compared to newer JDK.) Support for running tests is done for all systems. This includes adding code to run tests on windows x86, which was missing. Currently only jdk_tier1 on linuxes and macos is enabled, as currently only this is passing without failures (see details lower). Testing on windows is currently disabled (no tier1 groups are passing there currently). Disabled tests can be easily enabled once, they are fixed. Ultimate goal is of course to run all tier1 groups (jdk_tier1, langtools_tier1, hotspot_tier1) on all platforms.

**Tier1 status:**

jdk_tier1
- enabled on linuxes and macos (passes there)
- disabled on on windows, where `com/sun/jdi/*.sh` shell tests are currently failing

langtools_tier1
- disabled everywhere
- 1 failure: `tools/javac/diags/CheckExamples.java` (all platforms, see: [1])

hotspot_tier1
- disabled everywhere
- there are several faiures of `compiler/rtm/*` tests. These tests seem to broken on some systems (see: [2]), including machines used by GHA. Some tests are already excluded [3], but there are few more. Also exclusions were made only on x64 but it also affects x86.
- on x86 (32-bit) platforms there are few additional failures

[1] https://bugs.openjdk.org/browse/JDK-8265527
[2] https://bugs.openjdk.org/browse/JDK-8183263
[3] https://bugs.openjdk.org/browse/JDK-8226899

---------
### Progress

### Issue
 * [JDK-8294863](https://bugs.openjdk.org/browse/JDK-8294863): Enable partial tier1 testing in GHA for JDK8

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev pull/131/head:pull/131` \
`$ git checkout pull/131`

`$ git checkout pull/131` \
`$ git pull https://git.openjdk.org/jdk8u-dev pull/131/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 131`

`$ git pr show -t 131`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/131.diff">https://git.openjdk.org/jdk8u-dev/pull/131.diff</a>

</details>

### DIFF
--- a/.github/workflows/submit.yml
+++ b/.github/workflows/submit.yml
@@ -178,18 +178,20 @@ jobs:
         run: make CONF_NAME=linux-x64 LOG_LEVEL=debug images
         working-directory: jdk
 
+      - name: Pack artifacts
+        run: |
+          tar -cf jdk-${{ env.JDK_VERSION }}-internal+0_linux-x64_bin${{ matrix.artifact }}.tar.gz -C jdk/build/linux-x64/images j2sdk-image
+
       - name: Persist test bundles
         uses: actions/upload-artifact@v2
         with:
           name: transient_jdk-linux-x64${{ matrix.artifact }}_${{ needs.prerequisites.outputs.bundle_id }}
           path: |
-            jdk/build/linux-x64/bundles/jdk-${{ env.JDK_VERSION }}-internal+0_linux-x64_bin${{ matrix.artifact }}.tar.gz
-            jdk/build/linux-x64/bundles/jdk-${{ env.JDK_VERSION }}-internal+0_linux-x64_bin-tests${{ matrix.artifact }}.tar.gz
+            jdk-${{ env.JDK_VERSION }}-internal+0_linux-x64_bin${{ matrix.artifact }}.tar.gz
 
   linux_x64_test:
     name: Linux x64
     runs-on: "ubuntu-20.04"
-    if: false
     needs:
       - prerequisites
       - linux_x64_build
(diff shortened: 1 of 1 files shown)


zzambers:
@jerboaa, thanks, GHA are currently running on branch, from which PR was made, in my repo. I hope, they will then show up here. (If not, I am not sure what to do to make them show up.)

zzambers:
I think that was maybe some glitch by GH. Results were not showing up at first, later started to (somewhat randomly) appear. However now it seems they are displaying correctly.

jerboaa (approved):
Thanks! This is an improvement to the current status quo (no tests via GHA). No risk for product code as it's only changing `.github` directory files. Looking forward to getting more tests enabled.

zzambers:
@jerboaa thanks

gnu-andrew:
Hmm, just seen this. I'm not keen on how much this is diverging from the other JDKs. We'll likely need to revert some of this when backporting bundle support.

Do we not need a test bundle for running the tests? I had a feeling that was the case, as 8u doesn't have GTest, but wasn't 100%

zzambers:
> Hmm, just seen this. I'm not keen on how much this is diverging from the other JDKs. We'll likely need to revert some of this when backporting bundle support.

Sorry if I was interfering with something, you are working on. Just wanted to have tier1 testing enabled in GHA as it makes backporting much more convenient. If you do some backports, making JDK8 closer to newer JDKs, feel free to revert (update) parts of my changes concerning this matter.

> Do we not need a test bundle for running the tests? I had a feeling that was the case, as 8u doesn't have GTest, but wasn't 100%
 
When it comes to test bundle/image support, I am not sure there is currently use for this in JDK8 (tests in their current form). As far as I understand it, test image is mostly for tests, which make use of JNI. Native test libraries in newer JDKs are then built by build system and path to built libraries is then passed to jtreg using -nativepath argument [1]. Test image allows to (pre)build these, so they don't have to be built at test time.

However from what I can tell, JDK8 does not (yet) use this approach (building native test libraries by built system in makefiles). Instead tests which make use of JNI are implemented as shell tests and they build native libraries as part of test run. [2] (Other tests using JNI mostly follow similar boilerplate, with some modifications)

When testing GHA is concerned, I don't think, tests using JNI are causing any big problems. You just need working C/C++ compiler at test time. This is not really an issue especially when all Windows tests using JNI seem excluded [3]. Only problem I have seen was multilib configuration (32-bit JDK on 64-bit linux). While some tests can handle this case correctly using appropriate C/C++ compiler arguments [4], few have a problem of building 64bit JNI library for 32bit JDK and fail. There are only 3 tests affected by this in tier1 (concretely in hotspot_tier1) and they should be easy enough to fix. 
```
compiler/criticalnatives/argumentcorruption/Test8167409.sh 
runtime/jni/CallWithJNIWeak/test.sh 
runtime/jni/ReturnJNIWeak/test.sh 
```
So while newer JDK probably use cleaner approach for building native test libraries, I don't see it as a high priority in JDK8 (At least when it comes to GHA).

[1] https://github.com/openjdk/jdk/blob/ae60599e2ba75d80c3b4279903137b2c549f8066/make/RunTests.gmk#L801
[2] https://github.com/openjdk/jdk8u-dev/blob/5b5e548fba1073a1f87df3c71f0d09b63e033ced/hotspot/test/runtime/jni/CallWithJNIWeak/test.sh#L71
[3] https://github.com/openjdk/jdk8u-dev/blob/5b5e548fba1073a1f87df3c71f0d09b63e033ced/hotspot/test/runtime/jni/CallWithJNIWeak/test.sh#L60
[4] https://github.com/openjdk/jdk8u-dev/blob/5b5e548fba1073a1f87df3c71f0d09b63e033ced/hotspot/test/runtime/6929067/Test6929067.sh#L42

PR #154, opened 2022-11-03
8269039: Disable SHA-1 Signed JARs

alexeybakhtin:
I'd like to backport this enhancement for parity with Oracle and the security roadmap [1]

The patch is based on the OpenJDK11 backport [2]

The changes are the following:

1. java.security.* files are changed on the base of java.security
2. jdk/src/share/classes/sun/security/tools/jarsigner/Main.java : signJar method is merged manually because of differences in the JDK-8172404 backport. 
3. jdk/test/java/security/Security/signedfirst/DynStatic.java : modules dependency removed, changed path to utility classes
4. jdk/test/sun/security/tools/jarsigner/Test4431684.java : changed library path. List.of() is not available in JDK8, so it is replaced with Arrays.asList()
5. jdk/test/lib/security/SecurityUtils.java is updated to make removeFromDisabledAlgs method public. It is required by newly added test Test4431684.java
6. jdk/test/sun/security/tools/jarsigner/DefaultOptions.java is skipped, it was introduced in JDK9 by JDK-8049834 as default_options.sh but never backported to JDK8
7. JDK8 has jdk/test/sun/security/tools/jarsigner/diffend.sh test instead of jdk/test/sun/security/tools/jarsigner/DiffEnd.java. diffend.sh was not renamed to DiffEnd.java because of JDK-8180573 is not backported to JDK8. JDK-8180573 is a big refactoring and out of scope for this issue. diffend.sh updated accordingly - SHA1 replaced to SHA-256
8. JDK8 has jdk/test/sun/security/tools/jarsigner/ec.sh test instead of jdk/test/sun/security/tools/jarsigner/EC.java. ec.sh was not renamed to EC.java because of JDK-8180573 is not backported to JDK8. JDK-8180573 is a big refactoring and out of scope for this issue. ec.sh has all required changes by JDK-8172404
9. JDK8 has jdk/test/sun/security/tools/jarsigner/nameclash.sh test instead of jdk/test/sun/security/tools/jarsigner/NameClash.java. nameclash.sh was not renamed to NameClash.java because of JDK-8180573 is not backported to JDK8. JDK-8180573 is a big refactoring and out of scope for this issue. nameclash.sh has all required changes by JDK-8172404
10. JDK8 has jdk/test/sun/security/tools/jarsigner/oldsig.sh test instead of  jdk/test/sun/security/tools/jarsigner/OldSig.java. oldsig.sh was not renamed to OldSig.java because of JDK-8180573 is not backported to JDK8. JDK-8180573 is a big refactoring and out of scope for this issue. The changes in the oldsig.sh are not required because of JDK-8217375 is not backported to JDK8.
11. jdk/test/sun/security/tools/jarsigner/OldSig.props is not backported as it is not used in the  jdk/test/sun/security/tools/jarsigner/oldsig.sh

All java/security/Security sun/security/tools regression tests passed

[1] - https://www.java.com/en/jre-jdk-cryptoroadmap.html
[2] - https://github.com/openjdk/jdk11u-dev/commit/5a0824ba813ceda47847c9162c8a10bb0b8898e8

---------
### Progress

### Issue
 * [JDK-8269039](https://bugs.openjdk.org/browse/JDK-8269039): Disable SHA-1 Signed JARs

### Reviewers
 * [Martin Balao](https://openjdk.org/census#mbalao) (@martinuy - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev pull/154/head:pull/154` \
`$ git checkout pull/154`

`$ git checkout pull/154` \
`$ git pull https://git.openjdk.org/jdk8u-dev pull/154/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 154`

`$ git pr show -t 154`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/154.diff">https://git.openjdk.org/jdk8u-dev/pull/154.diff</a>

</details>


jerboaa:
@martinuy Could you please help review this? Thank you!

jdk/test/java/security/Security/signedfirst/DynStatic.java
```diff
@@ -0,0 +1,93 @@
...
+ * @library /lib /lib/testlibrary
+ * @run main/othervm DynStatic
+ */
+
+import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+
+import jdk.test.lib.process.ProcessTools;
```
martinuy, line 36: The `ProcessTools` class in JDK-8 has always been `jdk.testlibrary.ProcessTools`. To use it, you would only need `@library /lib/testlibrary` in the test's JTREG header. Unfortunately, one backport -which I presume to be JFR- introduced the same class, from a newer JDK release, at a different location. So we now have duplicated code for this test library and it is causing confusion: I can now see a couple of TLS test using this library when they shouldn't. Until we get rid of this technical debt, I suggest that all JDK-8 tests keep using `jdk.testlibrary.ProcessTools`.

jdk/test/sun/security/tools/jarsigner/TimestampCheck.java
```diff
@@ -422,29 +436,27 @@ public static void main(String[] args) throws Throwable {
...
+                        .shouldHaveExitValue(4)
+                        .shouldContain("jar signed")
+                        .shouldContain("with signer errors")
+                        .shouldMatch("SHA-1.*-tsadigestalg.*is disabled");
+                verify("sha1tsaalg.jar", "-verbose")
+                        .shouldHaveExitValue(16)
+                        .shouldContain("treated as unsigned")
+                        .shouldMatch("Timestamp.*digest.*SHA-1.*(disabled)");
 
                 // Disabled algorithms
```
martinuy, line 461: Isn't this comment removed?

martinuy (commented):
Hi @alexeybakhtin ,

Thanks for proposing this backport.

In addition to the inlined comments, I have this additional observation:

 * `jdk/test/sun/security/tools/jarsigner/oldsig.sh` is using SHA1, as well as JDK-11's OldSig.java. The security properties file that is now passed has the following comment: "Re-enable SHA-1 since OldSig.java test uses it". You mentioned 8217375 as the reason why changes are not required for the backport, but I'm having difficulties to find the connection and wonder if we should re-enable SHA1 for `oldsig.sh`. Can you please elaborate a bit on this?

Thanks,
Martin.-

alexeybakhtin:
Hello @martinuy

jdk/test/java/security/Security/signedfirst/DynStatic.java
alexeybakhtin: Thank you. You are right. jdk.testlibrary.ProcessTools is a better choice even if some modifications are required.

jdk/test/sun/security/tools/jarsigner/TimestampCheck.java
alexeybakhtin: Thank you. Fixed

martinuy: Good, thanks.

jdk/test/java/security/Security/signedfirst/DynStatic.java
martinuy: Good, thanks.

martinuy:
I've verified that test `oldsig.sh` passes with and without SHA1. This is because even if a disabled algorithm is used for signing or if a disabled algorithm is found when verifying a signature, `jarsigner` returns `0`. While the jar is actually signed, the verification fails and the jar is considered unsigned. The test should do better in the assertion statement, for example by checking `sm  ... B.class` in a verbose output. The test in later JDK releases has been fixed and the assertion statement improved.

@alexeybakhtin, my suggestion would be to backport the JDK-11 OldSig.java test to JDK-8. I agree with you that doing all the .sh -> .Java test conversions is out of the scope of this backport, but I would make an exception for the case discussed here because, otherwise, we would be having a broken/useless test in JDK-8. I'm also open consider adding a better assertion statement to the current .sh test. What do you think?

alexeybakhtin:
@martinuy, thank you for the review again.
I've updated the existing oldsig.sh and added OldSig.props file. Now test validates the signature.

martinuy:
@alexeybakhtin , thanks for addressing this concern.

What do you think about the following (minor) change to your proposal?

diff --git a/jdk/test/sun/security/tools/jarsigner/oldsig.sh b/jdk/test/sun/security/tools/jarsigner/oldsig.sh
index fcc2293d5d..5c8e328367 100644
--- a/jdk/test/sun/security/tools/jarsigner/oldsig.sh
+++ b/jdk/test/sun/security/tools/jarsigner/oldsig.sh
@@ -79,7 +79,8 @@ ${TESTJAVA}${FS}bin${FS}jarsigner \
     -digestalg SHA1 \
     B.jar c
 
-echo "${TESTJAVA}${FS}bin${FS}jarsigner -verify ${KS_ARGS} ${PROP_ARGS} -verbose B.jar"
-LINES=\`${TESTJAVA}${FS}bin${FS}jarsigner -verify ${KS_ARGS} ${PROP_ARGS} -verbose B.jar | grep smk | grep B.class | wc -l\`
+JAR_VERIFY_CMD="${TESTJAVA}${FS}bin${FS}jarsigner -verify ${KS_ARGS} ${PROP_ARGS} -verbose B.jar"
+echo ${JAR_VERIFY_CMD}
+LINES=\`${JAR_VERIFY_CMD} | grep smk | grep B.class | wc -l\`
 [ $LINES = 1 ] || exit 1

alexeybakhtin:
@martinuy, Thank you! Sure it is better. Test updated

martinuy (approved):
Looks good to me. Thanks for this contribution.

jerboaa:
The [crypto roadmap](https://www.java.com/en/jre-jdk-cryptoroadmap.html) mentions it's been included in Oracle JDK 8u351 (October 2022 CPU). While not 100% backwards compatible it's the *default* that is changing, right? So we should have it in OpenJDK 8u too and the [CSR mentions](https://bugs.openjdk.org/browse/JDK-8264362) it can be worked around by modifying `java.security` file. Is my understanding correct?

alexeybakhtin:
Right, it can be easily workarounded by removing SHA1 restriction from java.security file

alexeybakhtin:
> This is scary stuff. Clearly it isn't a backwards-compatible change. I guess the way this works is that JARs timestamped prior to January 01, 2019 are accepted, but only until the signing certificate expires. Right?

The behavior of the SHA-1 signed certificates before January 01, 2019 is not changed. They are still valid even if the signer certificate expires. The changes apply to JARs signed after January 01, 2019

martinuy:
Notice that for the 2019-01-01 max date to work, the JAR has to be timestamped by a TSA (Time Stamp Authority).

While the default behavior is changed (there is a [CSR attached to this bug](https://bugs.openjdk.org/browse/JDK-8272155) that has to be proposed for 8u as I commented before), the user can restore backward compatibility by re-enabling the use of the SHA-1 algorithm without conditions for JAR signing. You can verify how this works in the fix for [this test](https://github.com/openjdk/jdk8u-dev/pull/154/commits/57d0b0ba5a93c5ca2a88b2d8abc7f6f24a56a853). Also, the Release Note attached to this bug indicates how to [restore backward compatibility](https://bugs.openjdk.org/browse/JDK-8274081).

These crypto changes are part of the [crypto roadmap](https://www.java.com/en/jre-jdk-cryptoroadmap.html) as @jerboaa  mentioned. I think it's good to be aligned on that roadmap for parity with other JDKs and for security. The change has been backported to other JDK releases too, such as 11u. While I understand the concerns discussed here, I think that we have countermeasures to mitigate that and my recommendation to 8u maintainers is to approve this change.

gnu-andrew:
> Notice that for the 2019-01-01 max date to work, the JAR has to be timestamped by a TSA (Time Stamp Authority).
> 
> While the default behavior is changed (there is a [CSR attached to this bug](https://bugs.openjdk.org/browse/JDK-8272155) that has to be proposed for 8u as I commented before), the user can restore backward compatibility by re-enabling the use of the SHA-1 algorithm without conditions for JAR signing. You can verify how this works in the fix for [this test](https://github.com/openjdk/jdk8u-dev/pull/154/commits/57d0b0ba5a93c5ca2a88b2d8abc7f6f24a56a853). Also, the Release Note attached to this bug indicates how to [restore backward compatibility](https://bugs.openjdk.org/browse/JDK-8274081).
> 
> These crypto changes are part of the [crypto roadmap](https://www.java.com/en/jre-jdk-cryptoroadmap.html) as @jerboaa mentioned. I think it's good to be aligned on that roadmap for parity with other JDKs and for security. The change has been backported to other JDK releases too, such as 11u. While I understand the concerns discussed here, I think that we have countermeasures to mitigate that and my recommendation to 8u maintainers is to approve this change.

I agree it needs to be done, but I think, at this stage, it should have been delayed to 8u372. There are follow-on issues for this - e.g. https://bugs.openjdk.org/browse/JDK-8275887 - which will  now have to be resolved during rampdown.

alexeybakhtin:
@gnu-andrew, Thank you for the comment. I've submitted [PR](https://github.com/openjdk/jdk8u-dev/pull/197 ) for the backport of JDK-8275887

gnu-andrew:
Thanks. It'll need to be against https://github.com/openjdk/jdk8u once I've finished promoting b05 to make 8u362. jdk8u-dev is now for 8u372 changes.

alexeybakhtin:
@gnu-andrew, New pull request against jdk8u was created : https://github.com/openjdk/jdk8u/pull/24

PR #182, opened 2022-11-16
8241086: Test runtime/NMT/HugeArenaTracking.java is failing on 32bit Windows

zzambers:
Test `hotspot/test/runtime/NMT/HugeArenaTracking.java` (from hotspot/tier1) currently fails on Windows x86. This backport fixes this test to only run on 64-bit archs.

---------
### Progress

### Issue
 * [JDK-8241086](https://bugs.openjdk.org/browse/JDK-8241086): Test runtime/NMT/HugeArenaTracking.java is failing on 32bit Windows

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev pull/182/head:pull/182` \
`$ git checkout pull/182`

`$ git checkout pull/182` \
`$ git pull https://git.openjdk.org/jdk8u-dev pull/182/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 182`

`$ git pr show -t 182`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/182.diff">https://git.openjdk.org/jdk8u-dev/pull/182.diff</a>

</details>

### DIFF
--- a/hotspot/test/runtime/NMT/HugeArenaTracking.java
+++ b/hotspot/test/runtime/NMT/HugeArenaTracking.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, Red Hat, Inc. All rights reserved.
+ * Copyright (c) 2019, 2020, Red Hat, Inc. All rights reserved.
  *
  * This code is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License version 2 only, as
@@ -25,6 +25,7 @@
  * @test
  * @key nmt jcmd
  * @library /testlibrary /testlibrary/whitebox
+ * @requires vm.bits == 64
  * @build HugeArenaTracking
  * @run main ClassFileInstaller sun.hotspot.WhiteBox
  * @run main/othervm -Xbootclasspath/a:. -XX:+UnlockDiagnosticVMOptions -XX:+WhiteBoxAPI -XX:NativeMemoryTracking=detail HugeArenaTracking
--- a/hotspot/test/runtime/NMT/HugeArenaTracking.java
+++ b/hotspot/test/runtime/NMT/HugeArenaTracking.java
@@ -25,7 +25,6 @@
  * @test
  * @key nmt jcmd
  * @library /testlibrary /testlibrary/whitebox
- * @requires vm.bits == 64
  * @build HugeArenaTracking
  * @run main ClassFileInstaller sun.hotspot.WhiteBox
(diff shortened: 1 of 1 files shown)


hotspot/test/runtime/NMT/HugeArenaTracking.java
```diff
@@ -25,6 +25,7 @@
  * @test
  * @key nmt jcmd
  * @library /testlibrary /testlibrary/whitebox
+ * @requires vm.bits == 64
```
jerboaa, line 28: I don't think this has any effect on JDK 8. AFAIK, `vm.bits` is being set in JDK 11+ via `test/jtreg-ext/requires/VMProps.java` which isn't available in JDK 8. Have you tested this works?

I think something like this would be suitable here:

```
if (!Platform.is64Bit()) {
   System.out.println("Test requires 64 bits. Skipping...");
   return;
}
```

zzambers: Oh, you are right. I ran tests with 32-bit OSes enabled [1] and failure there disappeared, so I considered it working. I haven't realized test was also excluded on 64-bits. (condition was basically treated as always false) Nice find, I'll fix it. Thanks

[1] https://github.com/zzambers/jdk8u-dev/actions/runs/3471605048/jobs/5802167187

zzambers: But that reminds me, when I doing this backport, I  was actually wondering if it is supported on 8, so I grepped jdk8u-dev repo to see if there is some other tests doing this. There is one [2]. And it is indeed also skipped everywhere. So one more test to be fixed was discovered... :)

[2] https://github.com/openjdk/jdk8u-dev/blob/7ae002caba9ef17db5c105f76478424ce4cde04a/hotspot/test/runtime/Metaspace/MaxMetaspaceSizeTest.java#L29

zzambers:
Fixed. Tested with 32-bit testing enabled.  Passes. (Skipped on 32-bits, no longer skipped on 64-bits.)

https://github.com/zzambers/jdk8u-dev/actions/runs/3482910658

jerboaa (approved):
LGTM.

zzambers:
@jerboaa thanks for the review
/integrate

zzambers:
Oh, it does not yet have jdk8-fix-yes (will need before sponsor)

PR #252, opened 2023-01-30
8301332: [8u] Fix writing of test files after the cgroups v2 backport

jerboaa:
Please review this trivial refactoring of test files. It fixes usages of `Files.write(..., String.getBytes())`  to `Files.write(..., String.getBytes(StandardCharsets.UTF-8))`. Affected files don't contain non-ascii content, but it's a good cleanup to do anyway.

Testing:
[x] Tests compile, run and pass (as before)

Thoughts?


gnu-andrew (dismissed):
Thanks for fixing these cases. I know it's likely low risk, but I have also seen cases in the past where the fix has to be to make the character set or language explicit, and these are the kind of issues you don't see until someone runs the test with an unusual default character set.

jerboaa:
Thanks for the review and approval!

gnu-andrew (approved):

PR #254, opened 2023-01-31
8301550: [8u] Enable additional linux build testing in GitHub

zzambers:
Parts of testing in GitHub are currently still disabled as they needed changes to work with JDK 8.

This change does fixes for JDK 8 and enables them. Enabled tests include building hotspot in different configurations and also cross building for other architectures.

**Notes on modifications:**
- `optimized` jvm-variant does not exist on JDK 8, `client` variant tested instead
- added support for multilib testing to linux additional (to existing native and cross) as `client` and `minimal` variants are only for 32-bits on JDK 8
- arm and s390x are only available as `zero` variant on JDK 8 so they use this variant (added required libffi to buildroot)
- JDK 8 does not support `--with-build-jdk` configuration option (not to be confused with boot jdk), so it was removed as well as code preparing build jdk
- dropped dependence of linux additional on linux x64 build (was only needed to get build jdk)
- In case of JDK 8, a bit more persuasion was required to successfully configure cross build. Explicit configuration of libs was required. Also  *FLAGS env. variables with sysroot had to be set for configuration to succeed, but they must not remain set for build, otherwise it fails.

**Testing:**
All test passed in Github CI.

---------
### Progress

### Issue
 * [JDK-8301550](https://bugs.openjdk.org/browse/JDK-8301550): [8u] Enable additional linux build testing in GitHub

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev pull/254/head:pull/254` \
`$ git checkout pull/254`

`$ git checkout pull/254` \
`$ git pull https://git.openjdk.org/jdk8u-dev pull/254/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 254`

`$ git pr show -t 254`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/254.diff">https://git.openjdk.org/jdk8u-dev/pull/254.diff</a>

</details>

### DIFF
--- a/.github/workflows/submit.yml
+++ b/.github/workflows/submit.yml
@@ -10,7 +10,7 @@ on:
       platforms:
         description: "Platform(s) to execute on"
         required: true
-        default: "Linux x64, Linux x86, Windows x64, Windows x86, macOS x64"
+        default: "Linux additional (hotspot only), Linux x64, Linux x86, Windows x64, Windows x86, macOS x64"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -23,7 +23,7 @@ jobs:
     outputs:
       should_run: ${{ steps.check_submit.outputs.should_run }}
       bundle_id: ${{ steps.check_bundle_id.outputs.bundle_id }}
-      platform_linux_additional: false
+      platform_linux_additional: ${{ steps.check_platforms.outputs.platform_linux_additional }}
       platform_linux_x64: ${{ steps.check_platforms.outputs.platform_linux_x64 }}
       platform_linux_x86: ${{ steps.check_platforms.outputs.platform_linux_x86 }}
       platform_windows_x64: ${{ steps.check_platforms.outputs.platform_windows_x64 }}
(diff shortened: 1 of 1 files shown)


gnu-andrew (approved):
Looks good to me.

When I did the bare minimum to get the builds working, these additional builds and testing had to be turned off due to the need to pass a JDK bundle between steps, which 8u didn't (and still doesn't) support. It looks like for this, we didn't even need it anyway.

I haven't dug into `--with-build-jdk` in the later JDKs a lot, but I believe the idea is to save rebuilding the platform independent class files. So our 8u build will take a bit longer as we'll be rebuilding those in cross-compilation, but at least we're not trying to run javac on a Zero-based architecture.

It will be good to have these extra architectures tested, especially arm[32] as I don't believe that's in any of my testing on RHEL and Fedora any more.  Please add `jdk8u-fix-request` to the bug and I'll approve.

zzambers:
@gnu-andrew Thanks for the review

zzambers:
> Hmmm, it seems the additional build is quite quick because it only does `make hotspot`. We should maybe consider changing that to `make images`, as with the other targets, in a later PR, so that the native JDK code is also tested.

Yy, I have not changed it and kept it hotspot only. I have not (yet) tried to do full build, but hopefully it should work. Can be done with later PR, as you said, if desired.

zzambers:
@gnu-andrew, thanks
/integrate

PR #235, opened 2023-01-24
8301143: [TESTBUG] jfr/event/sampling/TestNative was backported to JDK8u without proper native wrapper

judovana:
This test was brought in by 8223147: JFR Backport and never passed. Despite not being wrong nor having jfr broken in jdk8. jdk8 testsuite simply lacks native support cases.

This is adding the shell wrapper which is compiling and properly setting up the native longSleep  in native shared object.

This commit was tested only on linux.
After this commit whole jdk_jfr group passes

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8301143](https://bugs.openjdk.org/browse/JDK-8301143): [TESTBUG] jfr/event/sampling/TestNative was backported to JDK8u without proper native wrapper

### Reviewers
 * [Severin Gehwolf](https://openjdk.org/census#sgehwolf) (@jerboaa - **Reviewer**)
 * [Paul Hohensee](https://openjdk.org/census#phh) (@phohensee - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev pull/235/head:pull/235` \
`$ git checkout pull/235`

Update a local copy of the PR: \
`$ git checkout pull/235` \
`$ git pull https://git.openjdk.org/jdk8u-dev pull/235/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 235`

View PR using the GUI difftool: \
`$ git pr show -t 235`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk8u-dev/pull/235.diff">https://git.openjdk.org/jdk8u-dev/pull/235.diff</a>

</details>

PR #234, opened 2023-01-19
8282947: JFR: Dump on shutdown live-locks in some conditions

judovana:
Reviewed-by: mgronlun

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8282947](https://bugs.openjdk.org/browse/JDK-8282947): JFR: Dump on shutdown live-locks in some conditions

### Reviewers
 * [Paul Hohensee](https://openjdk.org/census#phh) (@phohensee - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev pull/234/head:pull/234` \
`$ git checkout pull/234`

Update a local copy of the PR: \
`$ git checkout pull/234` \
`$ git pull https://git.openjdk.org/jdk8u-dev pull/234/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 234`

View PR using the GUI difftool: \
`$ git pr show -t 234`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk8u-dev/pull/234.diff">https://git.openjdk.org/jdk8u-dev/pull/234.diff</a>

</details>

PR #255, opened 2023-01-31
8287463: JFR: Disable TestDevNull.java on Windows

judovana:
---------
### Progress

### Issue
 * [JDK-8287463](https://bugs.openjdk.org/browse/JDK-8287463): JFR: Disable TestDevNull.java on Windows

### Reviewers
 * [Severin Gehwolf](https://openjdk.org/census#sgehwolf) (@jerboaa - **Reviewer**) ⚠️ Review applies to [4e1ca73c](https://git.openjdk.org/jdk8u-dev/pull/255/files/4e1ca73ccb304b407532ed7c5f4818c84fc724fd)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev pull/255/head:pull/255` \
`$ git checkout pull/255`

`$ git checkout pull/255` \
`$ git pull https://git.openjdk.org/jdk8u-dev pull/255/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 255`

`$ git pr show -t 255`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/255.diff">https://git.openjdk.org/jdk8u-dev/pull/255.diff</a>

</details>

### DIFF
--- a/jdk/test/jdk/jfr/api/recording/dump/TestDumpDevNull.java
+++ b/jdk/test/jdk/jfr/api/recording/dump/TestDumpDevNull.java
@@ -32,7 +32,7 @@ import jdk.jfr.Recording;
  * @summary Tests that it's possible to dump to /dev/null without a livelock
  * @key jfr
  *
- * @library /lib / /testlibrary
+ * @library /lib /
  * @run main/othervm jdk.jfr.api.recording.dump.TestDumpDevNull
  */
 public class TestDumpDevNull {
--- a/jdk/test/jdk/jfr/api/recording/dump/TestDumpDevNull.java
+++ b/jdk/test/jdk/jfr/api/recording/dump/TestDumpDevNull.java
@@ -38,6 +38,11 @@ import jdk.jfr.Recording;
 public class TestDumpDevNull {
 
     public static void main(String[] args) throws Exception {
+        String os = System.getProperty("os.name").toLowerCase();
+        if (os.startsWith("win")) {
+          System.out.println("Skipping TestDumpDevNull on windows: it uses /dev/null which is not present on windows");
+          System.exit(0);
+        }
         try (Recording r1 = new Recording()) {
             r1.setDestination(new File("/dev/null").toPath());
             r1.start();
--- a/jdk/test/jdk/jfr/api/recording/dump/TestDumpDevNull.java
+++ b/jdk/test/jdk/jfr/api/recording/dump/TestDumpDevNull.java
@@ -32,14 +32,13 @@ import jdk.jfr.Recording;
  * @summary Tests that it's possible to dump to /dev/null without a livelock
  * @key jfr
  *
- * @library /lib /
+ * @library /lib / /testlibrary
  * @run main/othervm jdk.jfr.api.recording.dump.TestDumpDevNull
  */
 public class TestDumpDevNull {
 
     public static void main(String[] args) throws Exception {
-        String os = System.getProperty("os.name").toLowerCase();
-        if (os.startsWith("win")) {
+        if (jdk.testlibrary.Platform.isWindows()) {
           System.out.println("Skipping TestDumpDevNull on windows: it uses /dev/null which is not present on windows");
           System.exit(0);
         }
--- a/jdk/test/jdk/jfr/api/recording/dump/TestDumpDevNull.java
+++ b/jdk/test/jdk/jfr/api/recording/dump/TestDumpDevNull.java
@@ -27,6 +27,8 @@ import java.io.File;
 
 import jdk.jfr.Recording;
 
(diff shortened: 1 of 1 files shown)


jerboaa:
@judovana It's targeting `master` branch, while I think it should target `pr/234`. The effect is that changes part of #234 don't show up in this PR. See https://mail.openjdk.org/pipermail/jdk-dev/2021-March/005232.html

judovana:
Thanx a lot for sharing the link. Updated

jerboaa (approved):
Looks OK to me.

judovana:
> @judovana btw. this can also be done by Platform class (from test library):
> 
> https://github.com/openjdk/jdk8u-dev/blob/165ce6aa0bab08e9fbad3ac2495295278d3cf5c6/jdk/test/lib/testlibrary/jdk/testlibrary/Platform.java#L44
> 
> e.g.:
> 
> https://github.com/openjdk/jdk8u-dev/blob/7d3c0bede34930cadd76644e58bf56f2a83c3d01/hotspot/test/runtime/memory/ReadFromNoaccessArea.java#L62
> 
> but probably does not matter here (Platform class literally does the same thing :) )

I will humbly stay library-less.  It had bittenme a bit in the https://github.com/openjdk/jdk8u-dev/pull/235

jerboaa:
Right. I forgot about `Platform`. Yes, please use it. If we need to change how windows is being detected we only need to change one place not many.

judovana:
Sure. If somebody ever manage to make
```
  String os = System.getProperty("os.name").toLowerCase();
        if (os.startsWith("win")) {
...
```
not working. This test will be of least concern:)

judovana:
Hm. It do not work. The library is not found. It seems to me, that in jdk_jfr  target is unable to locate it no metter of what I do. Can we go back to  to property?

zzambers:
@judovana for Platform class I linked it would need to be `@library /lib/testlibrary`, I think. Path is relative to `TEST.ROOT`. I accidentally posted example from hotspot, which has it's copy of that class placed a bit differently. Note that there is also another copy of whole test library in jdk (including Platform class), which came with jfr backport [1]. It is placed directly in `/lib` and uses package `jdk.test.lib`. This is a one you currently switched to. Tests in JDK 8 are unfortunately a bit messy.

[1] https://github.com/openjdk/jdk8u-dev/commit/df7e09043392d5952d522a28702c6e5ec3e8492e

kvergizova:
@judovana maybe it's better to use `@requires (os.family != "windows")` similar to the original test and not change the code of the test? As I can see it works fine for jdk8 as well.

judovana:
> @judovana for Platform class I linked it would need to be `@library /lib/testlibrary`, I think. Path is relative to `TEST.ROOT`. I accidentally posted example from hotspot, which has it's copy of that class placed a bit differently. Note that there is also another copy of whole test library in jdk (including Platform class), which came with jfr backport [1]. It is placed directly in `/lib` and uses package `jdk.test.lib`. This is a one you currently switched to. Tests in JDK 8 are unfortunately a bit messy.
> 
> [1] [df7e090](https://github.com/openjdk/jdk8u-dev/commit/df7e09043392d5952d522a28702c6e5ec3e8492e)

Indeed. the test now passes. Although I was originally not aware of the copy of library in hotspot.

judovana:
> @judovana maybe it's better to use `@requires (os.family != "windows")` similar to the original test and not change the code of the test? As I can see it works fine for jdk8 as well.

The `@requires` comes from jtreg library. Alhough you can supply newer jtreg library, the minimal version enforced by jdl8 had not supported @requires; but after additional reading I think it already does support @requires. Will return the os.family as in original fix.

judovana:
Seems like it is working as expected: https://github.com/judovana/jdk-tester/actions/runs/4056269325/jobs/7016661070#step:12:187

judovana:
78 tests in jdk8 have `@requires os...`
Also I remeber earlier usage of `@requires`  was working as before  "removed @requires vm.hasJFR "  (https://github.com/openjdk/jdk8u-dev/pull/234/commits/eca929f50213385681d60510d5c14e0173eafb1e )  jtreg correctly complained, that it do not understand the `vm.hasJFR` so @requires was processed deep enoug.

zzambers:
Looked into `os.family` and this one indeed seems to work ok (there are tests using it in tier1 and they seem to be correctly selected based on OS). Note however that `@requires` requires some caution on JDK8 as not everything works. There was a problem with `vm.bits`, which has shown no errors, but tests got skipped on all platforms (even when supposed to be enabled)  [1].

[1] https://github.com/openjdk/jdk8u-dev/pull/222

jerboaa (approved):
OK

PR #269, opened 2023-02-27
8298108: Add a regression test for JDK-8297684

jerboaa:
Please review this backport of the regression test for a critical fix which we've included with the January CPU update. The fix was https://bugs.openjdk.org/browse/JDK-8280890

This patch adds an additional regression test which was seen in the wild with JDK 17. The patch isn't clean since the test library in jdk8u is different so needed a bit of a different setup routine for getting the jar with the provider created. Additionally, in jdk 8 the property to add providers is setting `java.ext.dirs=...` over `java.locale.providers=SPI` in later JDKs.

Testing: The test passes for me on the current dev tree and fails with the fix from JDK-8280890 reverted.

Thoughts?

---------
### Progress

### Issues
 * [JDK-8298108](https://bugs.openjdk.org/browse/JDK-8298108): Add a regression test for JDK-8297684
 * [JDK-8298271](https://bugs.openjdk.org/browse/JDK-8298271): java/security/SignedJar/spi-calendar-provider/TestSPISigned.java failing on Windows

### Reviewers
 * [Paul Hohensee](https://openjdk.org/census#phh) (@phohensee - **Reviewer**) ⚠️ Review applies to [2e937ca2](https://git.openjdk.org/jdk8u-dev/pull/269/files/2e937ca2e1508acf39e56fb6a4836dac7f8679f9)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/269/head:pull/269` \
`$ git checkout pull/269`

`$ git checkout pull/269` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/269/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 269`

`$ git pr show -t 269`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/269.diff">https://git.openjdk.org/jdk8u-dev/pull/269.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk8u-dev/pull/269#issuecomment-1446944011)

### DIFF
--- a/jdk/test/java/security/SignedJar/spi-calendar-provider/TestSPISigned.java
+++ b/jdk/test/java/security/SignedJar/spi-calendar-provider/TestSPISigned.java
@@ -105,7 +105,7 @@ public class TestSPISigned {
             // run test, which must not throw a NPE
             List<String> testRun = new ArrayList<>();
             String extDir = System.getProperty("java.home") + File.separator + "lib" + File.separator + "ext";
-            testRun.add("-Djava.ext.dirs=" + System.getProperty("java.home") + extDir + File.pathSeparator + MODS_DIR.toAbsolutePath().toString());
+            testRun.add("-Djava.ext.dirs=" + extDir + File.pathSeparator + MODS_DIR.toAbsolutePath().toString());
             testRun.add("-cp");
             String classPath = System.getProperty("java.class.path");
             testRun.add(classPath);
--- a/jdk/test/java/security/SignedJar/spi-calendar-provider/TestSPISigned.java
+++ b/jdk/test/java/security/SignedJar/spi-calendar-provider/TestSPISigned.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright (c) 2022, Red Hat, Inc.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+import jdk.testlibrary.JarUtils;
+import jdk.testlibrary.SecurityTools;
+import jdk.testlibrary.Asserts;
+import jdk.testlibrary.ProcessTools;
+import jdk.testlibrary.OutputAnalyzer;
+import java.util.Calendar;
+import java.util.Locale;
+import java.util.List;
+import java.util.ArrayList;
+import java.nio.file.Paths;
+import java.nio.file.Path;
+import java.nio.file.Files;
+import java.io.File;
+import static java.util.Calendar.WEDNESDAY;
+
+/*
+ * @test
+ * @bug 8297684 8269039
+ * @summary Checking custom CalendarDataProvider with SPI contained in signed jar does
+ *          not produce NPE.
+ * @library /lib/testlibrary
+ * @library provider
+ * @build baz.CalendarDataProviderImpl
+ * @run main/timeout=600 TestSPISigned
+ */
+public class TestSPISigned {
+
+    private static final String TEST_CLASSES = System.getProperty("test.classes", ".");
+    private static final String TEST_SRC = System.getProperty("test.src", ".");
+
+    private static final Path META_INF_DIR = Paths.get(TEST_SRC, "provider", "meta");
+    private static final Path META_INF_DIR_SOURCE = META_INF_DIR.resolve("META-INF");
+    private static final Path SERVICES_DIR_SOURCE = META_INF_DIR_SOURCE.resolve("services");
+    private static final Path PROVIDER_SOURCE = SERVICES_DIR_SOURCE.resolve("java.util.spi.CalendarDataProvider");
+    private static final Path PROVIDER_PARENT = Paths.get(TEST_CLASSES);
+    private static final Path PROVIDER_DIR = PROVIDER_PARENT.resolve("provider");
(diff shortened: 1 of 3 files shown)


phohensee (approved):
Lgtm.

jerboaa:
Thanks for the review, Paul!

jerboaa:
Please keep it open, bot.

jdk/test/java/security/SignedJar/spi-calendar-provider/TestSPISigned.java
```diff
@@ -0,0 +1,136 @@
...
+                                  "-storepass changeit -dname CN=test -alias test")
+                     .shouldHaveExitValue(0);
+            // sign jar
+            SecurityTools.jarsigner("-keystore ks -storepass changeit " +
+                                "-signedjar " + SIGNED_JAR + " " + UNSIGNED_JAR + " test")
+                     .shouldHaveExitValue(0);
+            // run test, which must not throw a NPE
+            List<String> testRun = new ArrayList<>();
+            String extDir = System.getProperty("java.home") + File.separator + "lib" + File.separator + "ext";
+            testRun.add("-Djava.ext.dirs=" + System.getProperty("java.home") + extDir + File.pathSeparator + MODS_DIR.toAbsolutePath().toString());
```
gnu-andrew, line 108: Doesn't this add `System.getProperty("java.home")` to the front of `extDir` which already begins with `System.getProperty("java.home")`?

jdk/test/java/security/SignedJar/spi-calendar-provider/TestSPISigned.java
```diff
@@ -0,0 +1,136 @@
...
+               throw new RuntimeException("Test failed! Expected 'run-test' arg for test run");
+            }
+        } else {
+            // Set up signed jar with custom calendar data provider
+            //
+            // 1. Create jar with custom CalendarDataProvider
+            Files.createDirectory(META_INF_DIR_TARGET);
+            Files.createDirectory(SERVICES_DIR_TARGET);
+            Files.copy(PROVIDER_SOURCE, PROVIDER_TARGET);
+            JarUtils.createJarFile(UNSIGNED_JAR, PROVIDER_DIR);
```
gnu-andrew, line 95: I take it the JAR file does not need to be updated with the META-INF directory (11u has `JarUtils.updateJarFile(UNSIGNED_JAR, META_INF_DIR);`) because it is instead located via `-Djava.ext.dirs`?

gnu-andrew (commented):
I've queried a couple of minor issues, but generally looks like a good conversion to 8u. It does seem that installing the provider in 8u is quite awkward. Is there no way to do it with just the JAR?

gnu-andrew:
We should also probably reference [JDK-8298108](https://bugs.openjdk.org/browse/JDK-8298108) in this PR to make it clear that the issue is also resolved for 8u (it was effectively "fixed" by not using a hardcoded path separator in the first place)

jdk/test/java/security/SignedJar/spi-calendar-provider/TestSPISigned.java
jerboaa: If you override the property, it will only contain what you pass it:

```
$ /usr/lib/jvm/java-1.8.0-openjdk/bin/java -Djava.ext.dirs=/foo -XshowSettings:properties -version 2>&1 | grep java.ext.dirs
    java.ext.dirs = /foo
$ /usr/lib/jvm/java-1.8.0-openjdk/bin/java -XshowSettings:properties -version 2>&1 | grep java.ext.dirs
    java.ext.dirs = /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.362.b09-2.fc37.x86_64/jre/lib/ext
```

The intention is to use the extra dir as well, but it should also use `${java.home}/ext`

jdk/test/java/security/SignedJar/spi-calendar-provider/TestSPISigned.java
jerboaa: This has to do with different test libraries in place in 8u. `JarUtils.createJarFile()` here creates a jar file based on a directory (in this case the `test.classes` dir), where we already have the `*.class` files and we copy the `META-INF` folder from the test sources there. The javadoc of `createJarFile()` has:

```
Creates a JAR file from the contents of a directory.

Equivalent to {@code jar cf <jarfile> -C <dir> .}
```

So we create the right jar file with the META-INF directory included right from the start (as compared to the 11u version, where a jar file is being created *without* `META-INF/services` files and then updated with `META-INF` bits).

jerboaa:
> It does seem that installing the provider in 8u is quite awkward. Is there no way to do it with just the JAR?

Agreed. I couldn't get it to work otherwise, though. The calendar providing jar needs to be in the `java.ext.dirs` list of directories **and** the jar have the service provider (via `ServiceLoader`).

jerboaa:
> We should also probably reference [JDK-8298271](https://bugs.openjdk.org/browse/JDK-8298271) in this PR to make it clear that the issue is also resolved for 8u (it was effectively "fixed" by not using a hardcoded path separator in the first place)

Done.

jdk/test/java/security/SignedJar/spi-calendar-provider/TestSPISigned.java
gnu-andrew: Yeah, I get that, but I'm talking about the simpler issue of including the value of `System.getProperty("java.home")` twice, once in `extDir` and then again in the string passed to `testRun.add`.

If I break out the code into a test file,

~~~
$ cat TestExt.java
import java.io.File;

import java.nio.file.Path;
import java.nio.file.Paths;

import java.util.ArrayList;
import java.util.List;

public class TestExt
{
    private static final Path MODS_DIR = Paths.get("mods");

	// run test, which must not throw a NPE
	List<String> testRun = new ArrayList<>();
	String extDir = System.getProperty("java.home") + File.separator + "lib" + File.separator + "ext";
	testRun.add("-Djava.ext.dirs=" + System.getProperty("java.home") + extDir + File.pathSeparator + MODS_DIR.toAbsolutePath().toString());
	System.out.println(testRun);
    }
}
$ $HOME/build/icedtea8/bin/java -version
openjdk version "1.8.0_362"
OpenJDK Runtime Environment (IcedTea 3.26.0) (Gentoo build 1.8.0_362-b09)
OpenJDK 64-Bit Server VM (build 25.362-b09, mixed mode)
$ $HOME/build/icedtea8/bin/javac TestExt.java 
$ $HOME/build/icedtea8/bin/java TestExt
[-Djava.ext.dirs=/home/andrew/build/icedtea8/jre/home/andrew/build/icedtea8/jre/lib/ext:/tmp/mods]
~~~

You can see the value passed includes `java.home` twice. I guess it's not causing an issue with the test because it doesn't actually need to use any of the extensions in the system extension directory.

jerboaa:
@gnu-andrew Is there anything else you need my action on?

jerboaa:
@gnu-andrew Ping? Could you please take another look. Thanks!

jerboaa:
Thanks for taking a look!

/integrate

PR #282, opened 2023-03-10
8278851: Correct signer logic for jars signed with multiple digest algorithms

jerboaa:
Please review this backport which fixes a regression new in 8u362 due to [JDK-8269039](https://bugs.openjdk.org/browse/JDK-8269039). For some jars which are signed with an obsolete digest algorithm, after JDK-8269039, they're treated as unsigned even though one of the signatories used a valid (non-obsolete) digest algo.

The backport is not clean, mainly because of JDK-8275887 (which in 11u wasn't present when that bug was backported), copyright headers and different test infra in 8u. So the test looks a bit different so that it works on 8u. All of the changes are fairly trivial to resolve. Changes in `SecurityUtils.java` aren't needed as 8u-dev already has them.

Testing:
 - [x] `jdk/test/sun/security/tools/jarsigner` tests pass.
 - [x] new regression test. Fails prior product fix, passes after.

---------
### Progress

### Issue
 * [JDK-8278851](https://bugs.openjdk.org/browse/JDK-8278851): Correct signer logic for jars signed with multiple digest algorithms

### Reviewers
 * [Martin Balao](https://openjdk.org/census#mbalao) (@martinuy - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/282/head:pull/282` \
`$ git checkout pull/282`

`$ git checkout pull/282` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/282/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 282`

`$ git pr show -t 282`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/282.diff">https://git.openjdk.org/jdk8u-dev/pull/282.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk8u-dev/pull/282#issuecomment-1464150018)

### DIFF
--- a/jdk/src/share/classes/java/util/jar/JarVerifier.java
+++ b/jdk/src/share/classes/java/util/jar/JarVerifier.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -96,6 +96,10 @@ class JarVerifier {
     /** collect -DIGEST-MANIFEST values for blacklist */
     private List<Object> manifestDigests;
 
+    /* A cache mapping code signers to the algorithms used to digest jar
+       entries, and whether or not the algorithms are permitted. */
+    private Map<CodeSigner[], Map<String, Boolean>> signersToAlgs;
+
     public JarVerifier(String name, byte rawBytes[]) {
         manifestName = name;
         manifestRawBytes = rawBytes;
@@ -105,6 +109,7 @@ class JarVerifier {
         pendingBlocks = new ArrayList<>();
         baos = new ByteArrayOutputStream();
         manifestDigests = new ArrayList<>();
+        signersToAlgs = new HashMap<>();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


jerboaa:
GHA failure on Linux x86 (32 bit) of `gc/logging/TestGCId.java` seems unrelated.

martinuy (approved):
Hi @jerboaa ,

Thanks for contributing this backport. Looks good to me. I've run the _JarWithOneNonDisabledDigestAlg_ test locally and passes.

Minor note —which is not worth any change to the current proposal, in my view—: the list that _List.of_ returns is unmodifiable, which is not the case for _Arrays.asList_.

Just out of curiosity, was there a reason for placing _JarWithOneNonDisabledDigestAlg_ into a new _disdigest_ directory? I see that in 11u and above is in the _jarsigner_ directory.

Thanks,
Martin.-

jerboaa:
Thanks for the review!

> Minor note —which is not worth any change to the current proposal, in my view—: the list that _List.of_ returns is unmodifiable, which is not the case for _Arrays.asList_.

OK. I could wrap it in `Collections.unmodifiableList(...)`, but it's probably not worth it in this case.
 
> Just out of curiosity, was there a reason for placing _JarWithOneNonDisabledDigestAlg_ into a new _disdigest_ directory? I see that in 11u and above is in the _jarsigner_ directory.

Yes, the test failed when adding to the `jarsigner` directory directly. At least when the full suite of tests was run in the `jarsigner` directory. If I remember correctly, this caused some weird CNFE.

gnu-andrew (approved):
8u backport looks correct. I had the same comment about `List.of` vs `asList` as Martin. It's not worth changing here, but, given how frequently this comes up, we should probably look at having a utility method to handle the conversion.

I see the lack of [JDK-8056174](https://bugs.openjdk.org/browse/JDK-8056174) results in having to rewrite the signJarFile method completely to call the external program, as there is no programmatic access to jar signing. Too risky to backport this just for a nicer test.

Approved for 8u

jerboaa:
Thanks!

/integrate

PR #301, opened 2023-04-11
Backport d30aeec30c3832b2514a6f06451897afdaf52fb3

tstuefe:
To reduce intermittent errors from TestPerfCountersAndMemoryPools, I'd like to backport "8151460: Metaspace counters can have inconsistent values".

This does not backport cleanly since JEP 280 [1] modified the test, and I won't backport that one of course. Therefore I also added the test-specific fix JEP 280 did (https://github.com/openjdk/jdk8u-dev/commit/24773b47512b522c9bfa346559a34138d900b77e).

Tests: I manually ran gc/metaspace jtreg tests on Linux x64 with fastdebug and release builds.

[1] https://bugs.openjdk.org/browse/JDK-8148483


tstuefe:
Thanks @jerboaa and @phohensee.

GHAs ran through, including the changed tests.

/integrate

PR #318, opened 2023-05-24
8193017: Import freetype sources into OpenJDK source tree

gdams:
Imports the Freetype source to be consistent with JDK11+

As discussed in https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/631 CC @gnu-andrew

---------
### Progress
- [ ] [JDK-8193017](https://bugs.openjdk.org/browse/JDK-8193017) needs maintainer approval

### Issue
 * [JDK-8193017](https://bugs.openjdk.org/browse/JDK-8193017): Import freetype sources into OpenJDK source tree (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/318/head:pull/318` \
`$ git checkout pull/318`

`$ git checkout pull/318` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/318/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 318`

`$ git pr show -t 318`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/318.diff">https://git.openjdk.org/jdk8u-dev/pull/318.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk8u-dev/pull/318#issuecomment-1560841172)


gdams:
Please keep this open

mrserb:
probably we should keep it open?

mrserb:
@gdams do you have a chance to look at the merge-conflict?

gdams:
Yup @mrserb I'll try and get to it this week. I must stress that it would be good to make a decision on this as it's very painful to keep such a large changeset up to date

gdams:
@mrserb I've resolved the merge conflicts. I'm slightly worried about the changes to `common/autoconf/generated-configure.sh` though. I've got `autoconf@2.69` on my mac (installed using brew) but there might be a more specific version required?

mrserb:
> @mrserb I've resolved the merge conflicts. I'm slightly worried about the changes to `common/autoconf/generated-configure.sh` though. I've got `autoconf@2.69` on my mac (installed using brew) but there might be a more specific version required?

I had a similar issue some time ago https://github.com/openjdk/jdk8u-dev/pull/413#issuecomment-1951600088, at that time I also run 2.69 on macos. Solved that by running autoconf on some linux hosts.

gdams:
> > @mrserb I've resolved the merge conflicts. I'm slightly worried about the changes to `common/autoconf/generated-configure.sh` though. I've got `autoconf@2.69` on my mac (installed using brew) but there might be a more specific version required?
> 
> I had a similar issue some time ago [#413 (comment)](https://github.com/openjdk/jdk8u-dev/pull/413#issuecomment-1951600088), at that time I also run 2.69 on macos. Solved that by running autoconf on some linux hosts.

that looks better, I've compiled it locally with the patch applied and the generated configure looks correct now

make/lib/Awt2dLibraries.gmk
```diff
@@ -0,0 +1,1130 @@
+#
```
mrserb, line 1: looks like this file duplicates the jdk/make/lib/Awt2dLibraries.gmk?

make/copy/Copy-java.desktop.gmk
```diff
@@ -0,0 +1,50 @@
...
+# questions.
+#
+
+include CopyCommon.gmk
+
+$(eval $(call IncludeCustomExtension, copy/Copy-java.desktop.gmk))
+
+################################################################################
+
+$(CONF_DST_DIR)/sound.properties: $(TOPDIR)/src/java.desktop/share/conf/sound.properties
```
mrserb, line 32: "java.desktop" module is used.

make/lib/Awt2dLibraries.gmk
gdams: correct, I'll remove it

make/copy/Copy-java.desktop.gmk
gdams: updated PTAL

gdams: Actually I wonder if this entire file is required at all, it doesn't appear to be used in JDK8

mrserb: right, I think all changes(the new files) in the "TOPDIR/make" folder are unnecessary

mrserb:
I did a quick test on macOS. If the "libfreetype.dylib" is deleted from some of the old jdk8 build(or jdk-21) the next exception is occurred for the font demo:

>./jdk-8/bin/java -jar ./build/macosx-x86_64-normal-server-release/images/j2sdk-image/demo/jfc/Font2DTest/Font2DTest.jar 
Exception in thread "main" java.lang.UnsatisfiedLinkError: /Users/openjdk/jdk8u-dev/jdk-8/jre/lib/libfontmanager.dylib: dlopen(/Users/openjdk/jdk8u-dev/jdk-8/jre/lib/libfontmanager.dylib, 0x0001): Library not loaded: @rpath/libfreetype.dylib.6
  Referenced from: <E45B410C-8F6D-3350-90ED-7D918CD703B6> /Users/openjdk/jdk8u-dev/jdk-8/jre/lib/libfontmanager.dylib
  Reason: tried: '/Users/openjdk/jdk8u-dev/jdk-8/jre/lib/./libfreetype.dylib.6' (no such file), '/Users/openjdk/jdk8u-dev/jdk-8/jre/lib/./libfreetype.dylib.6' (no such file), '/Users/openjdk/jdk8u-dev/jdk-8/jre/lib/server/./libfreetype.dylib.6' (no such file), '/Users/openjdk/jdk8u-dev/jdk-8/jre/lib/server/../libfreetype.dylib.6' (no such file), '/Users/openjdk/jdk8u-dev/jdk-8/bin/./libfreetype.dylib.6' (no such file), '/usr/lib/libfreetype.dylib.6' (no such file, not in dyld cache)
	at java.lang.ClassLoader$NativeLibrary.load(Native Method)
	at java.lang.ClassLoader.loadLibrary0(ClassLoader.java:1934)
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1838)
	at java.lang.Runtime.loadLibrary0(Runtime.java:871)
	at java.lang.System.loadLibrary(System.java:1124)
	at sun.lwawt.macosx.LWCToolkit$1.run(LWCToolkit.java:93)
	at sun.lwawt.macosx.LWCToolkit$1.run(LWCToolkit.java:80)
	at java.security.AccessController.doPrivileged(Native Method)
	at sun.lwawt.macosx.LWCToolkit.<clinit>(LWCToolkit.java:79)
....
	at java.awt.Toolkit.getDefaultToolkit(Toolkit.java:854)
....
	at Font2DTest.main(Font2DTest.java:1031)

But If I delete the "libfreetype.dylib" from the build after this patch the demo works fine.

mrserb:
> But If I delete the "libfreetype.dylib" from the build after this patch the demo works fine.

Using "DYLD_PRINT_LIBRARIES=YES" traced that the "Homebrew" version of freetype is loaded. it is unclear why this version is not loaded in case of jdk-21. @rpath looks the same.

mrserb:
> Using "DYLD_PRINT_LIBRARIES=YES" traced that the "Homebrew" version of freetype is loaded. it is unclear why this version is not loaded in case of jdk-21. @rpath looks the same.

nevermind seems that in case of "macos+Homebrew" the jdk-21 is affected as well but only if I build jdk locally.
tested on windows and works fine.

make/copy/Copy-java.desktop.gmk
gdams: agreed, I've removed all those changes

jdk/src/share/legal/freetype.md
```diff
@@ -0,0 +1,537 @@
+## The FreeType Project: Freetype v2.9
```
mrserb, line 1: The "legal" folder is used in jdk9+, for jdk8 licenses should be placed in THIRD_PARTY_README file. We have several of them in jdk8, but after https://github.com/openjdk/jdk8u-dev/pull/557 integration there will be only one.

common/autoconf/toolchain_windows.m4
```diff
@@ -152,12 +152,6 @@ AC_DEFUN([TOOLCHAIN_CHECK_POSSIBLE_VISUAL_STUDIO_ROOT],
 
       if test "x$VS_ENV_CMD" = x; then
         AC_MSG_NOTICE([Warning: None of $VCVARSFILES were found, Visual Studio installation not recognized. Ignoring])
-      else
-        # PLATFORM_TOOLSET is used during the compilation of the freetype sources
-        # (see 'LIB_BUILD_FREETYPE' in libraries.m4) and must be one of 'v100',
-        # 'v110' or 'v120' for VS 2010, 2012 or VS2013
-        eval PLATFORM_TOOLSET="\${VS_VS_PLATFORM_NAME_${VS_VERSION}}"
-
```
mrserb, line 160: this change probably correct, but it is not part of the jdk11 [commit](https://github.com/openjdk/jdk/commit/58ff4ee8c4ec399528338b59eadea996110366a0#diff-098656ab8708ea86ee2cc19047997ed8f58caf22b2d239630276329fae933269), and this code still exists in the [openjdk/jdk](https://github.com/openjdk/jdk/blob/74066bcca82749722e6fee57469520d418bf3430/make/autoconf/toolchain_microsoft.m4#L105)

doc/building.md
```diff
@@ -405,9 +405,6 @@ happens, read more below in [the `configure` options](#configureoptions).
 
 Some examples:
 
->  **Windows 32bit build with freetype specified:** \
```
mrserb, line 408: doc/building.html should be updated as well.

mrserb: note that the documentation for freetype in jdk8 is outdated, in jdk9 it was reworked by this [commit](https://github.com/openjdk/jdk/commit/8dc9c32d0d1bbc791555f231e51c55924f256e75). We should update at least the freetype related part as a follow-up patch.

common/autoconf/libraries.m4
```diff
@@ -253,123 +253,38 @@ AC_DEFUN_ONCE([LIB_SETUP_CUPS],
...
 AC_DEFUN([LIB_CHECK_POTENTIAL_FREETYPE],
 [
   POTENTIAL_FREETYPE_INCLUDE_PATH="$1"
   POTENTIAL_FREETYPE_LIB_PATH="$2"
   METHOD="$3"
 
-  # First check if the files exists.
-  if test -s "$POTENTIAL_FREETYPE_INCLUDE_PATH/ft2build.h"; then
-    # We found an arbitrary include file. That's a good sign.
+  # Assume we've found freetype to begin
```
mrserb, line 262: from where this comment is come from? the jdk11 and openjdk/jdk use "# Let's start with an optimistic view of the world :-)" added by the "https://github.com/openjdk/jdk/commit/6b1eac7c4e86e33cba1dd30d43b417675de9000f"

common/autoconf/libraries.m4
```diff
@@ -381,256 +296,168 @@ AC_DEFUN([LIB_CHECK_POTENTIAL_FREETYPE],
...
-      else
-        AC_MSG_WARN([--with-freetype-src is currently only supported on Windows - ignoring])
+  if (test "x$with_freetype_include" = "x" && test "x$with_freetype_lib" != "x") || \
+     (test "x$with_freetype_include" != "x" && test "x$with_freetype_lib" = "x"); then
+    AC_MSG_ERROR([Must specify both or neither of --with-freetype-include and --with-freetype-lib])
+  fi
+  FREETYPE_TO_USE=bundled
+  if test "x$OPENJDK_TARGET_OS" != "xwindows" && \
+      test "x$OPENJDK_TARGET_OS" != "xmacosx" && \
+      test "x$OPENJDK_TARGET_OS" != "xaix"; then
```
mrserb, line 328: I am not sure about "aix" do we want to use the bundled version on it?

mrserb:
it will be good to test it on solartis and aix. @yan please take a look.

jdk/src/share/legal/freetype.md
gdams: I've copied the entire contents of this file to the THIRD_PARTY_README file. Hopefully that's the correct way to do it.

common/autoconf/libraries.m4
gdams: good spot, I've updated the comment but this is only a partial backport of those changes

common/autoconf/toolchain_windows.m4
gdams: good spot, I've pushed a commit to revert this and will see if a build still passes

theRealAph:
This one is outside the usual rules for 8u, which is closed for enhancements. For that reason, it must be accompanied by a _very_ strong justification, and evidence that it carries no risk.
By now, I'd have thought that everyone building JDK 8 on Windows etc. had workflows in place. We are at the (long) tail of the JDK 8 life cycle.

gnu-andrew:
> This one is outside the usual rules for 8u, which is closed for enhancements. For that reason, it must be accompanied by a _very_ strong justification, and evidence that it carries no risk. By now, I'd have thought that everyone building JDK 8 on Windows etc. had workflows in place. We are at the (long) tail of the JDK 8 life cycle.

I'm not sure I'd call this an enhancement. The lack of in-tree sources for FreeType means that there is no reference version of FreeType to build and test against, as there is with other libraries used by the AWT classes (libjpeg, libpng, giflib, LCMS). 

Historically, this is because Oracle did not use FreeType and its maintenance was largely left to others. There is evidence of this in this patch with the only JDK code changes being to allow FreeType to be used on non-OpenJDK builds (largely irrelevant to use I would expect)

Yes, those on Windows & Mac likely have some way of building against FreeType in place, but it is an ongoing burden to maintain that version of FreeType, and it means every person building OpenJDK potentially uses some different version of FreeType.

It's also an issue on the GNU/Linux side as there is no option *not* to use a system FreeType, which potentially reduces the portability of such builds. A build of OpenJDK may take place against one system installation of FreeType and then be used against a completely different one.

The reason these issues have not been a bigger problem up to now is FreeType is a pretty stable library. I think the risk of introducing this now is still very low (as I say, there are very few code changes at all) and it improves the maintainability of 8u for everyone.

doc/building.md
gnu-andrew: Yes, that commit probably needs to be backported as a starting point and then further changes made, including the other FreeType changes in the original version of this commit and replacing Mercurial with git.

I think this patch should at least update `doc/building.html` to match the `building.md` changes being made.

gnu-andrew (commented):
Patch looks good for comparison with the 11u version. My only real issue is with the doc updates. I thought at first we should update the other FreeType documentation here, but, after reading other comments, it seems better to follow up this work by trying to sync closer to later JDK versions first, then apply the changes from this changeset.  This patch should at least update `building.html` to match `building.md` though by regenerating the HTML.

I'll try bundled and system builds with this patch while you work on that.

gnu-andrew:
Keep open. Would be good to get this in.

mrserb:
Can we continue to work on this?

gnu-andrew:
Would also like to see this go in. Pending issues seem minor.

PR #339, opened 2023-07-12
8301119: Support for GB18030-2022

gnu-andrew:
This patch modifies GB18030 to handle both the 2000 and the 2022 variant. The 2000 variant is available by setting `-Djdk.charset.GB18030=2000`.  This PR replaces https://github.com/openjdk/jdk8u/pull/45.

With the preceding test changes in place (https://github.com/openjdk/jdk8u/pull/43 and https://github.com/openjdk/jdk8u/pull/44), the changes needed for this are fairly minimal. The biggest divergence from 11u is in the character set providers. The changes in the `make` directory are not needed as 8u never moved to using a template for GB18030 in the first place (the 11u changes revert it back to being source-based). The change in the SPI.java generator tool moves into ExtendedCharsets.java in the class library, as the file is not auto-generated in 8u. Following additional work by @jerboaa, the alias is now set to `2022` initially, and then replaced in the `init()` method if `jdk.charset.GB18030` is `2000`.

In 8u, the standard charsets are generated from a text file by a shell script, while the extended charsets are handled by a standard class. 11u moves GB18030 from extended to standard. I experimented with this in 8u, but it seemed more problematic than just keeping it in the extended set. The only reason I can see for moving it in 11u is it allowed `IS_2000` to be package-private to sun.nio.cs. This is resolved by @jerboaa's changes which modify `aliasMap` appropriately on creation, and so allow the use of a package-private method `isGB18030_2000` in `ExtendedCharsets` instead.

To use the 11u solution would mean major rewrites to the shell script or bringing over the whole change in how the standard charset provider is generated from 11u, which I think, along with moving the package the character set is in, is too risky and unnecessary for this change. The generation changes are necessary because the GB18030 character set needs to provide a different alias, depending on whether it is the 2000 or 2002 variant. The `genCharsetProvider.sh` would need the alterations we have added to `ExtendedCharsets.java` to handle this, but converted to `awk`. I did experiment with this, but saw test failures.

The only adjustment to the `GB18030.java` changes is copyright headers and the removal of `IS_2000` as mentioned above.

With the tests, the adjustments are just due to differing bug IDs, the absence of `@modules` and the use of constructs (`var`) and library calls (`Set.of`) that don't exist in 8u. The `List.of` and `Set.of` calls are frequent issues in backports, so I used this as an opportunity to introduce a full set of equivalents into the test library. It should now be possible to just rewrite `Set.of` to `Utils.setOf` and `List.of` to `Utils.listOf`. The returned collections are expected to be unmodifiable, not contain `null` and (in the case of sets) not contain duplicates. Simple replacement with a newly constructed `ArrayList` or `HashSet` would not ensure this. While this test does not rely on this, others may, so it seemed worth providing a closer replacement for use in future backports.

All `sun.nio.cs` and `java.nio.charset` tests pass with this patch applied.

---------
### Progress

### Issues
 * [JDK-8301119](https://bugs.openjdk.org/browse/JDK-8301119): Support for GB18030-2022 (**Enhancement** - P4)
 * [JDK-8310947](https://bugs.openjdk.org/browse/JDK-8310947): gb18030-2000 not selectable with LANG=zh_CN.GB18030 after JDK-8301119 (**Bug** - P3)

### Reviewers
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - Committer) ⚠️ Review applies to [dd4dccf6](https://git.openjdk.org/jdk8u-dev/pull/339/files/dd4dccf6ff8748ca2227e073c23602b50f7c8f53)

### Contributors
 * Severin Gehwolf `<sgehwolf@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/339/head:pull/339` \
`$ git checkout pull/339`

`$ git checkout pull/339` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/339/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 339`

`$ git pr show -t 339`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/339.diff">https://git.openjdk.org/jdk8u-dev/pull/339.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk8u-dev/pull/339#issuecomment-1632365188)


jdk/src/share/classes/sun/nio/cs/ext/ExtendedCharsets.java
```diff
@@ -1202,6 +1205,16 @@ protected void init() {
...
+        if (isGB18030_2000) {
+            deleteCharset("GB18030",
+                new String[] {
+                    "gb18030-2022"
+                });
+            charset("GB18030", "GB18030",
+                new String[] {
+                    "gb18030-2000"
+                });
+        }
```
tstuefe, line 1217: Can you please add this replacement to the comment above listing the changes done in init?

jdk/src/share/classes/sun/nio/cs/ext/ExtendedCharsets.java
```diff
@@ -1184,6 +1185,8 @@ protected void init() {
 
         String map = AccessController.doPrivileged(
             new GetPropertyAction("sun.nio.cs.map"));
+        boolean isGB18030_2000 = "2000".equals(AccessController.doPrivileged(
+                new GetPropertyAction("jdk.charset.GB18030")));
```
tstuefe, line 1189: More user friendly would be to test for "2000" or "2023" and to throw IAE for other values; but since later JDK versions don't do this either this is probably fine.

jdk/src/share/classes/sun/nio/cs/ext/GB18030.java
```diff
@@ -12595,6 +12637,7 @@ private CoderResult encodeArrayLoop(CharBuffer src,
             int condensedKey = 0;  // expands to a four byte sequence
             int hiByte = 0, loByte = 0;
             currentState = GB18030_DOUBLE_BYTE;
+            boolean isGB18030_2000 = ExtendedCharsets.isGB18030_2000();
```
tstuefe, line 12640: Here, and in other places: naming these vars IS_2000 would have simplified reviewing (mechanical comparison with 11)

jdk/test/lib/testlibrary/jdk/testlibrary/Utils.java
```diff
@@ -464,4 +467,782 @@ public static Path createTempDirectory(String prefix, FileAttribute<?>... attrs)
...
+     * @throws IllegalArgumentException if there are any duplicate elements
+     * @throws NullPointerException if an element is {@code null}
+     */
+    public static <E> Set<E> setOf(E e1, E e2, E e3, E e4, E e5, E e6) {
+        if (e1 == null) { throw new NullPointerException("e1"); }
+        if (e2 == null) { throw new NullPointerException("e2"); }
+        if (e3 == null) { throw new NullPointerException("e3"); }
+        if (e4 == null) { throw new NullPointerException("e4"); }
+        if (e5 == null) { throw new NullPointerException("e5"); }
+        if (e6 == null) { throw new NullPointerException("e6"); }
```
tstuefe, line 990: Here, and in other places: Objects.requireNonNull may be shorter.

jdk/test/lib/testlibrary/jdk/testlibrary/Utils.java
```diff
@@ -464,4 +467,782 @@ public static Path createTempDirectory(String prefix, FileAttribute<?>... attrs)
...
+        added = s.add(e2);
+        if (!added) { throw new IllegalArgumentException("duplicate 2"); }
+        added = s.add(e3);
+        if (!added) { throw new IllegalArgumentException("duplicate 3"); }
+        added = s.add(e4);
+        if (!added) { throw new IllegalArgumentException("duplicate 4"); }
+        added = s.add(e5);
+        if (!added) { throw new IllegalArgumentException("duplicate 5"); }
+        added = s.add(e6);
+        if (!added) { throw new IllegalArgumentException("duplicate 6"); }
```
tstuefe, line 1004: Could be shortened, but thats up to you (if one does not care for the string, a simple `if (!(s.add(e1) && s.add(e2) && ...` would be sufficient.

tstuefe (approved):
Hi Andrew,

I looked through the patch, and compared it with 11u, too, looks fine. Minor remarks inline, feel free to ignore the style nits.

jdk/src/share/classes/sun/nio/cs/ext/ExtendedCharsets.java
gnu-andrew: Good catch. I've updated the comment to document `jdk.charset.GB18030`.

jdk/src/share/classes/sun/nio/cs/ext/ExtendedCharsets.java
gnu-andrew: I agree, but yes this is something that should go into trunk and be backported down. I'm already a little uneasy about fixing 8310947 in 8u first.

jdk/src/share/classes/sun/nio/cs/ext/GB18030.java
gnu-andrew: I think some renaming is intentional as it's a static boolean constant in 11u, but a method and a private instance variable in 8u. I do think the `isGB18032_2000` is odd (presumably to avoid a name conflict with the method), so I will alter that to just `is_2000`.

gnu-andrew: Went with making that `isGB18030_2000` in the end and the method `IS_2000`. The latter certainly shortens some of the `GB18030.java` changes.

jdk/test/lib/testlibrary/jdk/testlibrary/Utils.java
gnu-andrew: Completely forgot that method existed!

jdk/test/lib/testlibrary/jdk/testlibrary/Utils.java
gnu-andrew: Not sure why I wrote this in such a long-winded way. Maybe there was more to it at one point and I didn't refine it after making further changes :/

Anyway, converted to a single line but kept the message, which can be useful in debugging (this is test output after all)

gnu-andrew:
Thanks for the quick review. I wasn't expecting you to be able to find time today.

I've made adjustments as discussed above. Tests pass as before.

tstuefe:
Still good. Thanks!

jerboaa (approved):
Looks fine to me.

gnu-andrew:
Thanks both. Flagged for approval with `jdk8u-fix-request`.

gnu-andrew:
I see `jdk8u-fix-yes`.
/integrate

PR #356, opened 2023-08-23
Backport 00e9c96d51bec53d4ae8a07c9c98af2c62f3d290

alexeybakhtin:
Backport applies almost cleanly.
Small merge fixes in the import section caused by JDK-8280890 (already applied in 8u) and JDK-8233228 (applied differently than in 11u)

No regression in jtreg tests

Follow-up JDK-8285696 fix should be applied


phohensee (approved):

jerboaa:
@alexeybakhtin This seems OK. Please create a PR with the follow-up so that both can be approved together. Thanks!

PR #361, opened 2023-08-30
Backport b77c161e7509aa3b09ebf3e6b2b1490c0667bbdc

simonis:
Hi all,

This pull request contains a backport of commit [b77c161e](https://github.com/openjdk/jdk11u-dev/commit/b77c161e7509aa3b09ebf3e6b2b1490c0667bbdc) from the [openjdk/jdk11u-dev](https://git.openjdk.org/jdk11u-dev) repository.

The commit being backported was authored by Volker Simonis on 30 Aug 2023 and was reviewed by Christoph Langer and Thomas Stuefe.

The backport applies cleanly except for the usual 11/8 directory shuffling and the following cosmetic context change in `jni.cpp`:
```patch
> --- a/jdk/src/share/native/com/sun/java/util/jar/pack/jni.cpp
> +++ b/jdk/src/share/native/com/sun/java/util/jar/pack/jni.cpp
58c58
<    CHECK_EXCEPTION_RETURN_VALUE(uPtr, 0);
---
>    CHECK_EXCEPTION_RETURN_VALUE(uPtr, NULL);
```
Thanks!

### DIFF
--- a/jdk/src/share/classes/com/sun/java/util/jar/pack/NativeUnpack.java
+++ b/jdk/src/share/classes/com/sun/java/util/jar/pack/NativeUnpack.java
@@ -60,7 +60,7 @@ class NativeUnpack {
 
     // Resets the engine and frees all resources.
     // Returns total number of bytes consumed by the engine.
-    private synchronized native long finish();
+    synchronized native long finish();
 
     // Setting state in the unpacker.
     protected  synchronized native boolean setOption(String opt, String value);
--- a/jdk/src/share/classes/com/sun/java/util/jar/pack/UnpackerImpl.java
+++ b/jdk/src/share/classes/com/sun/java/util/jar/pack/UnpackerImpl.java
@@ -139,6 +139,11 @@ public class UnpackerImpl extends TLGlobals implements Pack200.Unpacker {
                 } catch (UnsatisfiedLinkError | NoClassDefFoundError ex) {
                     // failover to java implementation
                     (new DoUnpack()).run(in0, out);
+                } finally {
+                    if (_nunp != null) {
+                       // Free up native memory and JNI handles to prevent leaks
+                       ((NativeUnpack) _nunp).finish();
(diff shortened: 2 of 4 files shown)


gnu-andrew (approved):
Backport looks ok (though not clean due to the context difference)

simonis:
Thanks @gnu-andrew! Regarding the "clean" label, I was a little confused by the following line in the [OpenJDK Backports Wiki](https://wiki.openjdk.org/display/SKARA/Backports):

> A backport commit is considered clean if the changes in the original commit are identical to the changes in the backport commit. Note that only the changes have to be identical, not the changed lines.

And I actually still don't fully understand what exactly "**_Note that only the changes have to be identical, not the changed lines_**" means?

gnu-andrew:
> Thanks @gnu-andrew! Regarding the "clean" label, I was a little confused by the following line in the [OpenJDK Backports Wiki](https://wiki.openjdk.org/display/SKARA/Backports):
> 
> > A backport commit is considered clean if the changes in the original commit are identical to the changes in the backport commit. Note that only the changes have to be identical, not the changed lines.
> 
> And I actually still don't fully understand what exactly "**_Note that only the changes have to be identical, not the changed lines_**" means?

It's the first time I've seen that myself and I agree it's confusing. I think the intention may be to imply that it's clean if you make the same code changes even if they are in a different place in the file or even a different file. To me, that would be going a bit far.

The approach I've always used, even before we used GitHub and Skara, is that a backport is clean if it can be automated by tools i.e. if you apply the shuffle script, and then patch, and it applies, it is clean. If you have to open a file and apply any hunks manually, that is not clean.

Even the former can be problematic if patch fuzzes the change into the wrong place, so I would be cautious even in those cases. That's why I actually quite like that 8u doesn't automatically add `clean` and patches do get a once over from a real human :) As someone has to approve anyway, I don't think it's really much extra work.

PR #374, opened 2023-09-19
6899049: G1: Clean up code in ptrQueue.[ch]pp and ptrQueue.inline.hpp

jianyesun:
Hi all,

This pull request contains a backport of commit [b256989eb34a32c8f03be448c0645baeb5192a01](https://github.com/openjdk/jdk11u-dev/commit/b256989eb34a32c8f03be448c0645baeb5192a01) from the [openjdk/jdk11u-dev](https://github.com/openjdk/jdk11u-dev) repository.

As reported by issue :  https://bugs.openjdk.org/browse/JDK-8316278 .  We found  the indexing method of PtrQueue's buf  is not  correct  when  converting an integer of type size_t to type int, then calling the method PtrQueue::byte_index_to_index . 
The key problem is this way of using: 
```
size_t i=0;    _buf[byte_index_to_index((int)i)] = NULL;  
```
The variable i of size_t type  cannot be converted directly to an int type . Other than that, the return value of the function byte_index_to_index is the index of the array _buf, and it should be non-negative. So it  should be a type of size_t. 
Currently we have found 2 issues related to this problem,  https://bugs.openjdk.org/browse/JDK-8308169 and  https://bugs.openjdk.org/browse/JDK-8303961.  They are all triggered by a special  size number of buf, like '-XX:G1UpdateBufferSize=512M'  or  '-XX:G1SATBBufferSize=500m'
We also added a test case.
Please review this PR. Thanks.

---------
### Progress
- [ ] [JDK-6899049](https://bugs.openjdk.org/browse/JDK-6899049) needs maintainer approval

### Issue
 * [JDK-6899049](https://bugs.openjdk.org/browse/JDK-6899049): G1: Clean up code in ptrQueue.[ch]pp and ptrQueue.inline.hpp (**Enhancement** - P5)

### Contributors
 * Yulong Liu `<liuyulong35@huawei.com>`
 * Kun Wang `<wangkun49@huawei.com>`
 * Jianye Sun `<sunjianye@huawei.com>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/374/head:pull/374` \
`$ git checkout pull/374`

`$ git checkout pull/374` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/374/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 374`

`$ git pr show -t 374`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/374.diff">https://git.openjdk.org/jdk8u-dev/pull/374.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk8u-dev/pull/374#issuecomment-1726077240)


jerboaa:
@jianyesun PR title needs to be `Backport b256989eb34a32c8f03be448c0645baeb5192a01` (provided the sha is correct).

jerboaa:
@jianyesun You are changing the title to something like `Backport b256989eb34a32c8f03be448c0645baeb5192a01 8316278: Fix the indexing method of PtrQueue's buf when a large integer value overflow`. It should be `Backport b256989eb34a32c8f03be448c0645baeb5192a01`.

The bots will do the rest.

jianyesun:
> @jianyesun You are changing the title to something like `Backport b256989eb34a32c8f03be448c0645baeb5192a01 8316278: Fix the indexing method of PtrQueue's buf when a large integer value overflow`. It should be `Backport b256989eb34a32c8f03be448c0645baeb5192a01`. I.e. drop `8316278: ...` from the title.
> 
> The bots will do the rest.

Finally done. Thank you for your patience and guidance,bro~

gnu-andrew:
Looks like the new test is failing.

gnu-andrew:
Where does this test come from? I don't see it in the patch being backported.

gnu-andrew:
> > Where does this test come from? I don't see it in the patch being backported.
> 
> We added it to check whether the problem is sovled. I saw there are errors like `Insufficient Memory Error` or `could not reserve enough space for 2097152KB object heap`. Maybe the resources of the CI environment are limited， the size of memory is less than 4G when the jvm starts with `-Xmx4096m`. When i change it to `-Xmx2048m`, it seems that the jvm cannot start or run with OOM.
> 
> Do you think it is necessary to add this test case ? If not, i will delete it. Or set the test not to execute in these three scenarios(Linux x86/windows x86/Windows x64) ?

Even with a working test, this is not the place to include something new unless it is specific to 8u.  If you do want to include it, it needs to be separated from your backport and proposed to https://github.com/openjdk/jdk under its own bug ID.  Once included there, it can be backported to 8u, as you have with JDK-6899049 here. Not only do we not want tests being unique to the 8u repository, but changes to the main repository get attention from those who are experts in this field. Stable update trees are generally expected to get fixes that have already been reviewed, but might need some minor modification to work on an older version, and so don't get the same reviewer coverage.

As to the test case itself, it's not clear to me what it's trying to test. Is it the command-line options? Or the actual allocation? I see other cases in the HotSpot tests where a 2GB heap is used, but they only run the VM with `-version` (e.g. https://github.com/openjdk/jdk/blob/HEAD/test/hotspot/jtreg/runtime/CompressedOops/UseCompressedOops.java ) 

The failure isn't architecture-specific, so the appropriate exclusion would be to check the available memory before running the VM process. If you only want to check the arguments work, then running with `-version` should be enough.

jianyesun:
> As to the test case itself, it's not clear to me what it's trying to test. Is it the command-line options? Or the actual allocation?

Well, as i described in issue [JDK-8316278](https://bugs.openjdk.org/browse/JDK-8316278) , the crash happends when looking for a element with a special index in PtrQueue's buf during  G1GC STAB processing.  Therefore, the value of G1SATBBufferSize is related to the size of the heap space or the size of the remaining heap space. We added this test which is already reported by other reporter (ie.see [JDK-8308169](https://bugs.openjdk.org/browse/JDK-8308169))  just to check whether this problem will still be triggered in linux x64. Maybe the test doesn't quite fit.

>Even with a working test, this is not the place to include something new unless it is specific to 8u. If you do want to include it, it needs to be separated from your backport and proposed to https://github.com/openjdk/jdk under its own bug ID. Once included there, it can be backported to 8u, as you have with JDK-6899049 here. Not only do we not want tests being unique to the 8u repository, but changes to the main repository get attention from those who are experts in this field. Stable update trees are generally expected to get fixes that have already been reviewed, but might need some minor modification to work on an older version, and so don't get the same reviewer coverage.

OK, i understand what you mean. Code's version management should really be considered. I have decided to delete the test.
By the way, this backport is clean. Indeed, the risk of index's type conversion is avoided. Therefore it is valuable. What do you think? 

We will continue to pay attention to the reasons why other platforms failed to pass in the future. Thanks.

hotspot/src/share/vm/gc_implementation/g1/dirtyCardQueue.cpp
```diff
@@ -291,10 +313,11 @@ void DirtyCardQueueSet::concatenate_logs() {
   for (JavaThread* t = Threads::first(); t; t = t->next()) {
     DirtyCardQueue& dcq = t->dirty_card_queue();
     if (dcq.size() != 0) {
-      void **buf = t->dirty_card_queue().get_buf();
+      void **buf = dcq.get_buf();
```
gnu-andrew, line 316: The 11u version changes this to `void** buf`. Was there a reason for the difference here?

hotspot/src/share/vm/gc_implementation/g1/ptrQueue.hpp
```diff
@@ -40,44 +40,49 @@ class PtrQueueSet;
...
+  bool is_permanent() const { return _permanent; }
 
   // Process queue entries and release resources, if not permanent.
   void flush_impl();
 
-public:
   // Initialize this queue to contain a null buffer, and be part of the
   // given PtrQueueSet.
-  PtrQueue(PtrQueueSet* qset, bool perm = false, bool active = false);
+  PtrQueue(PtrQueueSet* qset, bool permanent  = false, bool active = false);
```
gnu-andrew, line 79: An extra space has sneaked in here between `permanent` and the `=` sign.

gnu-andrew (changes requested):
A few minor unwanted differences have crept into the backport, but otherwise now looks mostly clean.

As to the test, if you do think it is worthwhile, please consider submitting it to openjdk/jdk.

hotspot/src/share/vm/gc_implementation/g1/dirtyCardQueue.cpp
jianyesun: I think the return type of get_buf which  is defined for the first time is inappropriate, see [here](https://github.com/openjdk/jdk8u-dev/blob/f4260a4deed5210fc56f08774e8ca50cf6f7cc8e/hotspot/src/share/vm/gc_implementation/g1/dirtyCardQueue.hpp#L74), first commited in [here](https://github.com/openjdk/jdk8u-dev/commit/18f3386a98e67f1c5b75c4cc47a0ebb898c59b9b). 
The 11u version changes the buf allocation by  `BufferNode::make_buffer_from_node`(see [here](https://github.com/openjdk/jdk11u/blob/master/src/hotspot/share/gc/g1/dirtyCardQueue.cpp#L178)), however the jdk8 does not do that.
I can modify it to `void** buf`, should  i do that?

hotspot/src/share/vm/gc_implementation/g1/ptrQueue.hpp
jianyesun: Thanks, i will fix it.

jianyesun:
Hi Andrew,  would you please review the code changes again?  Also, are there any other committers who can help review this MR? Please help me, thx~ @gnu-andrew

phohensee (changes requested):
White space mismatch in dirtyCardQueue.cpp, lines 209 and 210. dirtyCardQueue.hpp ine 72 also doesn't match, but Andrew allowed it so I will too. Other than that, lgtm.

jianyesun:
> White space mismatch in dirtyCardQueue.cpp, lines 209 and 210. dirtyCardQueue.hpp ine 72 also doesn't match, but Andrew allowed it so I will too. Other than that, lgtm.

Sorry, something interrupted me. I fix it as you suggested, thank you very much.

phohensee:
linux-x86 looks to have a couple of tier1 test failures. The windows-x64 failure looks like an infrastructure issue.

PR #432, opened 2024-01-26
8151255: javac with a annotation processor returns incorrect error code

liliBestCoder:
Hi,

When I was building my application using Maven 3.6.3, I had two applications A and B that had a dependency relationship. B relied on A's jar. When I built both A and B simultaneously, A built successfully. Although B threw an exception, Javac believed that B had built successfully. I unpacked B's  jar package and found that there was no class file inside. So I searched for the reason based on the exception information, and found that Javac was divided into multiple rounds when processing annotations. If an abort occurred in a certain round, Javac would not be able to find the delegate compiler. If Javac's own errors were 0 at this time, it would consider compilation successful, but te specific error information was reported to the global diagnosticCollector by the delegate compiler, So even though Maven printed an error, it didn't stop working due to the error, but instead continued to package and ultimately printed an incorrect package.

---------
### Progress
- [ ] [JDK-8151255](https://bugs.openjdk.org/browse/JDK-8151255) needs maintainer approval

### Issue
 * [JDK-8151255](https://bugs.openjdk.org/browse/JDK-8151255): javac with a annotation processor returns incorrect error code (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/432/head:pull/432` \
`$ git checkout pull/432`

`$ git checkout pull/432` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/432/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 432`

`$ git pr show -t 432`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/432.diff">https://git.openjdk.org/jdk8u-dev/pull/432.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk8u-dev/pull/432#issuecomment-1911442280)

### DIFF
--- a/langtools/src/share/classes/com/sun/tools/javac/main/JavaCompiler.java
+++ b/langtools/src/share/classes/com/sun/tools/javac/main/JavaCompiler.java
@@ -861,11 +861,29 @@ public class JavaCompiler {
             delegateCompiler.close();
             elapsed_msec = delegateCompiler.elapsed_msec;
         } catch (Abort ex) {
+            correctDelegateCompilerIfNeed();
             if (devVerbose)
                 ex.printStackTrace(System.err);
         } finally {
             if (procEnvImpl != null)
                 procEnvImpl.close();
+        }
+    }
+
+    /**
+     * If processAnnotations abort in a certain round,
+     * delegateCompiler is likely to be null,
+     * resulting in a compilation failure,
+     * but Javac considers it successful,
+     * so it may be necessary to find the final delegateCompiler compiler used here
+     */
+    private void correctDelegateCompilerIfNeed() {
+        if(delegateCompiler != null){
+            return;
+        }
+        Context procContext = procEnvImpl.getContext();
+        if(procContext != context){
+            delegateCompiler = JavaCompiler.instance(procContext);
         }
     }
 


jddarcy:
@liliBestCoder , per general OpenJDK development procedures, a fix like this should first be considered against the mainline release, currently JDK 23, and then if it is fixed in mainline, subsequently possibly backported to earlier release trains. You could also inquire to get feedback on a fix for a bug to the appropriate mailing list before sending the PR, compiler-dev in this case. HTH

liliBestCoder:
> @liliBestCoder , per general OpenJDK development procedures, a fix like this should first be considered against the mainline release, currently JDK 23, and then if it is fixed in mainline, subsequently possibly backported to earlier release trains. You could also inquire to get feedback on a fix for a bug to the appropriate mailing list before sending the PR, compiler-dev in this case. HTH

as been fixed in jdk 23. This should need to be tested to know. But I think jdk 23 should be very different from jdk 8. The patch should not be applied directly here. That's why I submitted PR here.

liliBestCoder:
> @liliBestCoder , per general OpenJDK development procedures, a fix like this should first be considered against the mainline release, currently JDK 23, and then if it is fixed in mainline, subsequently possibly backported to earlier release trains. You could also inquire to get feedback on a fix for a bug to the appropriate mailing list before sending the PR, compiler-dev in this case. HTH

I don't know if this problem has been fixed in jdk 23. This should need to be tested to know. But I think jdk 23 should be very different from jdk 8. The patch should not be applied directly here. That's why I submitted PR here

liliBestCoder:
I have looked at the code of jdk23 and found that the delegated compiler is no longer used when processing annotations, and when an abort occurs, the exception should be reported directly to the global diagnosticCollector, so this problem of mine no longer exists. Moreover, the code difference is relatively large and it is difficult to backport it to Java8.

liliBestCoder:
I have looked at the code of jdk23 and found that the delegated compiler is no longer used when processing annotations, and when an abort occurs, the exception should be reported directly to the global diagnosticCollector, so this problem of mine no longer exists. Moreover, the code difference is relatively large and it is difficult to backport it to Java8.

jddarcy:
> I have looked at the code of jdk23 and found that the delegated compiler is no longer used when processing annotations, and when an abort occurs, the exception should be reported directly to the global diagnosticCollector, so this problem of mine no longer exists. Moreover, the code difference is relatively large and it is difficult to backport it to Java8.

From a comment on the bug, the issue does not reproduce under JDK 9 so was presumably fixed as a side-effect of some JDK 9 javac change.

As a rule, OpenJDK generally follows a fix in later release and then backport model, so isolating the JDK 9 change the addressed this issue would, by default, be the route to fix it. However, the 8u maintainers are conservative in what fixes are allowed in, so they may decline to accept the fix even if it backported from a later release. HTH

liliBestCoder:
Oh, I saw that this problem was indeed solved in jdk9, and the relevant code is also reflected in jdk23. The versions of these codes are quite different from jdk8. So now that jdk8 has such a problem, do we need to isolate jdk9 to solve it? Or just ignore it?
Or let those using jdk8 solve this problem by upgrading the jdk version.

liliBestCoder:
> > I have looked at the code of jdk23 and found that the delegated compiler is no longer used when processing annotations, and when an abort occurs, the exception should be reported directly to the global diagnosticCollector, so this problem of mine no longer exists. Moreover, the code difference is relatively large and it is difficult to backport it to Java8.
> 
> From a comment on the bug, the issue does not reproduce under JDK 9 so was presumably fixed as a side-effect of some JDK 9 javac change.
> 
> As a rule, OpenJDK generally follows a fix in later release and then backport model, so isolating the JDK 9 change the addressed this issue would, by default, be the route to fix it. However, the 8u maintainers are conservative in what fixes are allowed in, so they may decline to accept the fix even if it backported from a later release. HTH

Oh, I saw that this problem was indeed solved in jdk9, and the relevant code is also reflected in jdk23. The versions of these codes are quite different from jdk8. So now that jdk8 has such a problem, do we need to isolate jdk9 to solve it? Or just ignore it?
Or let those using jdk8 solve this problem by upgrading the jdk version.

PR #449, opened 2024-02-20
Backport 704c02a180cafab1da03d5e5cfd09d92bc4cda8c

fthevenet:
Hi,

This is a backport of [JDK-8082592](https://bugs.openjdk.org/browse/JDK-8082592) which adds Structured Exception Handling (SEH) to guard 'JNI_CreateJavaVM' on Windows.

The motivation for this backport is to help reduce the number of occurences where a VM might crash will terminate VM immediately without writing an error log Windows, making root cause analysis very difficult in such cases.

In addition, I plan to also backport [JDK-8186199](https://bugs.openjdk.org/browse/JDK-8186199) as a follow-up, which does something similar to guard 'JNI_DestroyJavaVM'.

Thanks!

### DIFF
--- a/hotspot/src/cpu/x86/vm/vm_version_x86.cpp
+++ b/hotspot/src/cpu/x86/vm/vm_version_x86.cpp
@@ -390,15 +390,6 @@ class VM_Version_StubGenerator: public StubCodeGenerator {
   };
 };
 
-
-void VM_Version::get_cpu_info_wrapper() {
-  get_cpu_info_stub(&_cpuid_info);
-}
-
-#ifndef CALL_TEST_FUNC_WITH_WRAPPER_IF_NEEDED
-  #define CALL_TEST_FUNC_WITH_WRAPPER_IF_NEEDED(f) f()
-#endif
-
 void VM_Version::get_processor_features() {
 
   _cpu = 4; // 486 by default
@@ -412,9 +403,7 @@ void VM_Version::get_processor_features() {
   if (!Use486InstrsOnly) {
     // Get raw processor info
 
-    // Some platforms (like Win*) need a wrapper around here
-    // in order to properly handle SEGV for YMM registers test.
-    CALL_TEST_FUNC_WITH_WRAPPER_IF_NEEDED(get_cpu_info_wrapper);
+    get_cpu_info_stub(&_cpuid_info);
 
     assert_is_initialized();
     _cpu = extended_cpu_family();
--- a/hotspot/src/os/windows/vm/os_windows.cpp
+++ b/hotspot/src/os/windows/vm/os_windows.cpp
@@ -2870,17 +2870,6 @@ address os::win32::fast_jni_accessor_wrapper(BasicType type) {
 }
 #endif
 
-void os::win32::call_test_func_with_wrapper(void (*funcPtr)(void)) {
-  // Install a win32 structured exception handler around the test
-  // function call so the VM can generate an error dump if needed.
-  __try {
-    (*funcPtr)();
-  } __except(topLevelExceptionFilter(
-             (_EXCEPTION_POINTERS*)_exception_info())) {
-    // Nothing to do.
-  }
-}
-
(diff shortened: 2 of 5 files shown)


tstuefe:
Since the back port was not clean, could you describe what the delta is?

tstuefe:
Hi Frederic,

It looks okay.

If we have the time, to be sure, could you please do the following test: apply the following patch:

```
diff --git a/hotspot/src/share/vm/runtime/thread.cpp b/hotspot/src/share/vm/runtime/thread.cpp
index e6586c40cb..7fc4efbb0b 100644
--- a/hotspot/src/share/vm/runtime/thread.cpp
+++ b/hotspot/src/share/vm/runtime/thread.cpp
@@ -3329,6 +3329,8 @@ void Threads::threads_do(ThreadClosure* tc) {
   // If CompilerThreads ever become non-JavaThreads, add them here
 }
 
+volatile char* AAAA = nullptr;
+
 jint Threads::create_vm(JavaVMInitArgs* args, bool* canTryAgain) {
 
   extern void JDK_Version_init();
@@ -3503,6 +3505,11 @@ jint Threads::create_vm(JavaVMInitArgs* args, bool* canTryAgain) {
     ShouldNotReachHere();
   }
 
+
+  if (UseNewCode) {
+    (*AAAA) = 'A';
+  }
+
   // Always call even when there are not JVMTI environments yet, since environments
   // may be attached late and JVMTI must track phases of VM execution
   JvmtiExport::enter_start_phase();
```

and run this without and without this backport.

Then test this by running  `java -XX:+UnlockDiagnosticVMOptions -XX:+UseNewCode -version`

I expect:
- without the backport: a "silent-like" crash without a hs-err file
- with the backport: a crash with a hs-err file

Then we know for sure this works. Thank you!

gnu-andrew (approved):
Patch looks clean code-wise. The delta is mostly copyright headers (including `os_windows.hpp` already having a newer date thanks to JDK-8183925) and indentation. I do notice there is an additional change to the formatting of `HOTSPOT_JNI_CREATEJAVAVM_ENTRY` which makes it match the context of the patch. I don't see a problem with that.

gnu-andrew:
> Since the back port was not clean, could you describe what the delta is?

Yes, it is good form in these cases to describe what changes you had to make in the backport. Otherwise, the reviewers have to work this out for themselves and make assumptions as to why you made certain decisions.

tstuefe (approved):
Good. Thanks for the speedy backport work.

tstuefe:
BTW, @gnu-andrew, should we do something about the failing security tests that make everything red? Maybe at least exclude them?

fthevenet:
> > Since the back port was not clean, could you describe what the delta is?
> 
> Yes, it is good form in these cases to describe what changes you had to make in the backport. Otherwise, the reviewers have to work this out for themselves and make assumptions as to why you made certain decisions.

Conflicts arose from changes that were introduced in the target at a latter date and therefore not present in the original patch. I opted to keep things the same as they currently were and only add new lines from the patch, with three notable exceptions:
1. When the patch explicitly removed code that was present in the target
2. I picked the most recent copyright date (for some files that means the one from the target and others the one from the patch) though I'm not sure this is correct or if it matters much.
3. I used the target's indentation for `HOTSPOT_JNI_CREATEJAVAVM_ENTRY` rather that the source's because it felt less jarring.

fthevenet:
/approval request The motivation for this backport is to help reduce the number of occurrences where a VM might terminate immediately without writing an error log Windows, making root cause analysis very difficult in such cases.
Backport is not clean because of conflicting changes introduced  subsequently to the original patch.
Risk is low.

fthevenet:
To follow-up: I completed the test proposed by @tstuefe and I can confirm that:
* Provoking a crash **with** this backport results in an hs_err being written.
* Provoking a crash **without** this backport results in nothing but "Segmentation fault" being written to std_err and nothing more (e.g. no hs_err report).

PR #479, opened 2024-04-09
8329826: GCC 12 reports some compiler error when building jdk8

kuenking:
Env:
ldd (GNU libc) 2.38
gcc (GCC) 12.3.1 

Test the PR patch;
Build release/fastdebug/slowdebug version pass

Reported issue : https://bugs.openjdk.org/browse/JDK-8329826

---------
### Progress
- [ ] [JDK-8329826](https://bugs.openjdk.org/browse/JDK-8329826) needs maintainer approval

### Issue
 * [JDK-8329826](https://bugs.openjdk.org/browse/JDK-8329826): GCC 12 reports some compiler error when building jdk8 (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/479/head:pull/479` \
`$ git checkout pull/479`

`$ git checkout pull/479` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/479/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 479`

`$ git pr show -t 479`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/479.diff">https://git.openjdk.org/jdk8u-dev/pull/479.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jdk8u-dev/pull/479#issuecomment-2049119150)

### DIFF
--- a/hotspot/make/linux/makefiles/adlc.make
+++ b/hotspot/make/linux/makefiles/adlc.make
@@ -66,7 +66,7 @@ CXXFLAGS += -DASSERT
 
 # CFLAGS_WARN holds compiler options to suppress/enable warnings.
 # Compiler warnings are treated as errors
-CFLAGS_WARN = $(WARNINGS_ARE_ERRORS)
+CFLAGS_WARN = $(WARNINGS_ARE_ERRORS) -Wno-register
 CFLAGS += $(CFLAGS_WARN)
 
 OBJECTNAMES = \
--- a/hotspot/make/linux/makefiles/gcc.make
+++ b/hotspot/make/linux/makefiles/gcc.make
@@ -212,7 +212,7 @@ ifeq ($(USE_CLANG), true)
   WARNINGS_ARE_ERRORS += -Wno-return-type -Wno-empty-body
 endif
 
-WARNING_FLAGS = -Wpointer-arith -Wsign-compare -Wundef -Wunused-function -Wunused-value -Wformat=2 -Wreturn-type
+WARNING_FLAGS = -Wpointer-arith -Wsign-compare -Wundef -Wunused-function -Wunused-value -Wformat=2 -Wreturn-type -Wno-stringop-overflow
 
 ifeq ($(USE_CLANG),)
   # Since GCC 4.3, -Wconversion has changed its meanings to warn these implicit
--- a/hotspot/make/linux/makefiles/gcc.make
+++ b/hotspot/make/linux/makefiles/gcc.make
@@ -212,7 +212,7 @@ ifeq ($(USE_CLANG), true)
   WARNINGS_ARE_ERRORS += -Wno-return-type -Wno-empty-body
 endif
 
-WARNING_FLAGS = -Wpointer-arith -Wsign-compare -Wundef -Wunused-function -Wunused-value -Wformat=2 -Wreturn-type -Wno-stringop-overflow
+WARNING_FLAGS = -Wpointer-arith -Wsign-compare -Wundef -Wunused-function -Wunused-value -Wformat=2 -Wreturn-type
 
 ifeq ($(USE_CLANG),)
   # Since GCC 4.3, -Wconversion has changed its meanings to warn these implicit
--- a/hotspot/src/share/vm/opto/type.cpp
+++ b/hotspot/src/share/vm/opto/type.cpp
@@ -2552,16 +2552,18 @@ TypeOopPtr::TypeOopPtr(TYPES t, PTR ptr, ciKlass* k, bool xk, ciObject* o, int o
         } else if (klass() == ciEnv::current()->Class_klass() &&
                    _offset >= InstanceMirrorKlass::offset_of_static_fields()) {
           // Static fields
-          assert(o != NULL, "must be constant");
           ciField* field = NULL;
           if (o != NULL) {
             ciInstanceKlass* k = o->as_instance()->java_lang_Class_klass()->as_instance_klass();
             field = k->get_field_by_offset(_offset, true);
           }
-          assert(field != NULL, "missing field");
-          BasicType basic_elem_type = field->layout_type();
-          _is_ptr_to_narrowoop = UseCompressedOops && (basic_elem_type == T_OBJECT ||
-                                                       basic_elem_type == T_ARRAY);
+          if (field != NULL) {
+            BasicType basic_elem_type = field->layout_type();
+            _is_ptr_to_narrowoop = UseCompressedOops && is_reference_type(basic_elem_type);
+          } else {
+            // unsafe access
+            _is_ptr_to_narrowoop = UseCompressedOops;
+          }
         } else {
           // Instance fields which contains a compressed oop references.
           field = ik->get_field_by_offset(_offset, false);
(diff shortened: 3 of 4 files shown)


gnu-andrew:
The `register` issue is already being handled by the work ongoing in https://github.com/openjdk/jdk8u-dev/pull/357

I have not seen other failures. Are they specific to certain architectures?

We should not be introducing changes uniquely to 8u unless it is really necessary, so we need to look at this code in later JDKs.

kuenking:
> The `register` issue is already being handled by the work ongoing in #357
> 
> I have not seen other failures. Are they specific to certain architectures?
> 
> We should not be introducing changes uniquely to 8u unless it is really necessary, so we need to look at this code in later JDKs.



> The `register` issue is already being handled by the work ongoing in #357
> 
> I have not seen other failures. Are they specific to certain architectures?
> 
> We should not be introducing changes uniquely to 8u unless it is really necessary, so we need to look at this code in later JDKs.

kuenking:
> The `register` issue is already being handled by the work ongoing in #357
> 
> I have not seen other failures. Are they specific to certain architectures?
> 
> We should not be introducing changes uniquely to 8u unless it is really necessary, so we need to look at this code in later JDKs.

Thank you very much for your code review and also suggest.
I found this problem when using gcc12 under the aarch64 architecture. After referring to https://github.com/openjdk/jdk8u-dev/pull/357, I still encounter other problems. I may need to remove the changes in adlc.make, the others are still having build issues in gcc12.

I'll test it on x86 architecture, and I'll report back with the results.

hotspot/make/linux/makefiles/gcc.make
```diff
@@ -212,7 +212,7 @@ ifeq ($(USE_CLANG), true)
   WARNINGS_ARE_ERRORS += -Wno-return-type -Wno-empty-body
 endif
 
-WARNING_FLAGS = -Wpointer-arith -Wsign-compare -Wundef -Wunused-function -Wunused-value -Wformat=2 -Wreturn-type
+WARNING_FLAGS = -Wpointer-arith -Wsign-compare -Wundef -Wunused-function -Wunused-value -Wformat=2 -Wreturn-type -Wno-stringop-overflow
```
mirabilos, line 215: disabling the warning instead of fixing the code is not the proper way to deal with this

hotspot/make/linux/makefiles/adlc.make
```diff
@@ -66,7 +66,7 @@ CXXFLAGS += -DASSERT
 
 # CFLAGS_WARN holds compiler options to suppress/enable warnings.
 # Compiler warnings are treated as errors
-CFLAGS_WARN = $(WARNINGS_ARE_ERRORS)
+CFLAGS_WARN = $(WARNINGS_ARE_ERRORS) -Wno-register
```
mirabilos, line 69: that’s probably okay to ignore

hotspot/src/share/vm/opto/type.cpp
```diff
@@ -2553,8 +2553,11 @@ TypeOopPtr::TypeOopPtr(TYPES t, PTR ptr, ciKlass* k, bool xk, ciObject* o, int o
                    _offset >= InstanceMirrorKlass::offset_of_static_fields()) {
           // Static fields
           assert(o != NULL, "must be constant");
-          ciInstanceKlass* k = o->as_instance()->java_lang_Class_klass()->as_instance_klass();
-          ciField* field = k->get_field_by_offset(_offset, true);
+          ciField* field = NULL;
+          if (o != NULL) {
+            ciInstanceKlass* k = o->as_instance()->java_lang_Class_klass()->as_instance_klass();
+            field = k->get_field_by_offset(_offset, true);
+          }
```
mirabilos, line 2560: This is useless papering over, not a fix, and not even good because it makes the code less legible at no gain.

Better to convert the two `assert`s to actual checks which abort the control flow of the function if triggered.

mirabilos (changes requested):

hotspot/src/share/vm/opto/type.cpp
stooke: (disclaimer: I am not a reviewer)
This proposed change doesn't really handle the case where o == NULL, although it makes the compiler happy.  I propose something like:

    assert(o != NULL, "must be constant");
    if (o != NULL) {
        ciInstanceKlass* k = o->as_instance()->java_lang_Class_klass()->as_instance_klass();
        ciField* field = k->get_field_by_offset(_offset, true);
        assert(field != NULL, "missing field");
        BasicType basic_elem_type = field->layout_type();
        _is_ptr_to_narrowoop = UseCompressedOops && (basic_elem_type == T_OBJECT ||
                                                     basic_elem_type == T_ARRAY);
     } else {
        _is_ptr_to_narrowoop = (?? I am not familiar enough with this code to put a suggestion here)
    }

mirabilos: No, don’t `assert` *and* `if` something. (The compiler may delete one of these two under Undefined Behaviour rules.)

I’ve not looked at the callers, but returning an error or throwing an exception if `o == NULL` would be best. Do **not** rely on `assert`.

hotspot/make/linux/makefiles/gcc.make
kuenking: Thank you for your code review. 
The same solution is used in later JDK versions. https://bugs.openjdk.org/browse/JDK-8320212 ; 
So JDK8 can also use this solution.

hotspot/src/share/vm/opto/type.cpp
kuenking: This is a good suggestion, I will try to fix it, refer to https://github.com/openjdk/jdk/blob/ad78b7fa67ba30cab2e8f496e4c765be15deeca6/src/hotspot/share/opto/type.cpp#L3560C3-L3572

thanks.

hotspot/make/linux/makefiles/gcc.make
mirabilos: No, rather the other way round: please fix that in later JDK versions as well.

I’m only the Debian packager for JDK 8, so I have no relationship to the other versions. This doesn’t make my critique invalid.

kuenking: thanks.  Ok, I'll try to fix it forward through the code;
Since I'm not sure how many errors there are yet, this may take a while. I'll try;

kuenking: Hi mirabilos，
I looked up some information; reference: https://gcc.gnu.org/bugzilla/show_bug.cgi? ID=106297. I also found that the gcc12 in my test environment was not a pure version but a version after secondary development, which affected the misjudgment of the jdk8 compilation error.
I built and tested using https://ftp.gnu.org/gnu/gcc/. The -Werror=stringop-overflow= error does not exist, and hotspot/src/share/vm/gc_implementation/g1/concurrentMark.cpp does not need to be modified.
The verification is as follows:
image

Therefore, I think the changes in gcc.make and concurrentMark.cpp should be removed.
thanks;

zzambers:
Issue with register keyword has been fixed by:
https://github.com/openjdk/jdk8u-dev/pull/357

Issue in `type.cpp` seems like duplicate of these backports:
https://github.com/openjdk/jdk8u-dev/pull/536
https://github.com/openjdk/jdk8u-dev/pull/552
(which also include tests)

PR #544, opened 2024-07-09
8318039: GHA: Bump macOS and Xcode versions

zzambers:
This backport is written from scratch, as GHA workflows in jdk8u differ significantly from newer jdk versions.

Notes:
- additionally installed `gawk` to prevent build error on newer macos (see my [comment](https://github.com/openjdk/jdk17u-dev/pull/2206#issuecomment-2217722646))
- interestingly, unlike for newer jdks (11, 17), additional backports do not seem necessary to pass the build, probably build system of jdk8 is more relaxed, when it comes to warnings as errors
- fixes problem of stuck macos jobs in GHA (caused by dropping of macos-11 by GitHub)

Testing:
- macos builds pass
- there are some test failures on updated macos, probably problem in test environment, seem to be hostname related -> to be fixed (either here, or in separate PR)

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] [JDK-8318039](https://bugs.openjdk.org/browse/JDK-8318039) needs maintainer approval

### Issue
 * [JDK-8318039](https://bugs.openjdk.org/browse/JDK-8318039): GHA: Bump macOS and Xcode versions (**Enhancement** - P4 - Approved)

### Reviewers
 * [Severin Gehwolf](https://openjdk.org/census#sgehwolf) (@jerboaa - **Reviewer**)
 * [Andrew John Hughes](https://openjdk.org/census#andrew) (@gnu-andrew - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/544/head:pull/544` \
`$ git checkout pull/544`

Update a local copy of the PR: \
`$ git checkout pull/544` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/544/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 544`

View PR using the GUI difftool: \
`$ git pr show -t 544`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jdk8u-dev/pull/544.diff">https://git.openjdk.org/jdk8u-dev/pull/544.diff</a>

</details>

### Webrev
[...]

PR #553, opened 2024-07-25
8186787: clang-4.0 SIGSEGV in Unsafe_PutByte

zzambers:
This backport fixes [failures](https://github.com/zzambers/jdk8u-dev/actions/runs/9858657012/job/27221579883) (segfaults) in following tests which appeared after [macos update](https://github.com/openjdk/jdk8u-dev/pull/544):
```
sun/misc/CopyMemory.java 
compiler/unsafe/OpaqueAccesses.java
```
Backport differs from original changeset, because there were significant changes/refactoring in unsafe.

**Notes:**
- [original changeset](https://github.com/openjdk/jdk11u-dev/commit/6dc1d8c06d98e127b022886172e16b90bf357c97) changes pointer returned by `addr` (`MemoryAccess` class), to volatile. Otherwise it is basically just refactoring.
- `MemoryAccess` is used by `Unsafe_{Set,Put}*` and `Unsafe_{Set,Put}*Volatile` functions, defined using `DEFINE_GETSETOOP` and `DEFINE_GETSETOOP_VOLATILE` macros
- jdk8 does not have `MemoryAccess` class, so equivalent pointers, in functions mentioned higher, are cast to volatile, to achieve same effect

Testing:
Tier1: OK (fixes `sun/misc/CopyMemory.java` and `compiler/unsafe/OpaqueAccesses.java` tests on macos, 1 failure on Linux x86 is timeout - seems unrelated, macos failures explained here: https://github.com/openjdk/jdk8u-dev/pull/544#issuecomment-2250636257)

---------
### Progress
- [x] [JDK-8186787](https://bugs.openjdk.org/browse/JDK-8186787) needs maintainer approval

### Issue
 * [JDK-8186787](https://bugs.openjdk.org/browse/JDK-8186787): clang-4.0 SIGSEGV in Unsafe_PutByte (**Bug** - P2 - Approved)

### Reviewers
 * [Andrew Dinn](https://openjdk.org/census#adinn) (@adinn - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/553/head:pull/553` \
`$ git checkout pull/553`

`$ git checkout pull/553` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/553/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 553`

`$ git pr show -t 553`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/553.diff">https://git.openjdk.org/jdk8u-dev/pull/553.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jdk8u-dev/pull/553#issuecomment-2250618942)
</details>

### DIFF
--- a/hotspot/src/share/vm/prims/unsafe.cpp
+++ b/hotspot/src/share/vm/prims/unsafe.cpp
@@ -167,11 +167,11 @@ jint Unsafe_invocation_key_to_method_slot(jint key) {
 
 #define GET_FIELD(obj, offset, type_name, v) \
   oop p = JNIHandles::resolve(obj); \
-  type_name v = *(type_name*)index_oop_from_field_offset_long(p, offset)
+  type_name v = *(volatile type_name*)index_oop_from_field_offset_long(p, offset)
 
 #define SET_FIELD(obj, offset, type_name, x) \
   oop p = JNIHandles::resolve(obj); \
-  *(type_name*)index_oop_from_field_offset_long(p, offset) = truncate_##type_name(x)
+  *(volatile type_name*)index_oop_from_field_offset_long(p, offset) = truncate_##type_name(x)
 
 #define GET_FIELD_VOLATILE(obj, offset, type_name, v) \
   oop p = JNIHandles::resolve(obj); \
--- a/hotspot/src/share/vm/prims/unsafe.cpp
+++ b/hotspot/src/share/vm/prims/unsafe.cpp
@@ -165,6 +165,13 @@ jint Unsafe_invocation_key_to_method_slot(jint key) {
 #define truncate_jfloat(x) (x)
 #define truncate_jdouble(x) (x)
 
+// Note that GET_FIELD and SET_FIELD cast the object field's C type to
+// volatile at the point of access. That is needed in order to ensure
+// that the C compiler does not reorder the object field access wrt to
+// preceding and succeeding volatile accesses to the thread flag field
+// (made by UnsafeWrapper) which safeguard the field access. See
+// JDK-8186787 for details.
+
 #define GET_FIELD(obj, offset, type_name, v) \
   oop p = JNIHandles::resolve(obj); \
   type_name v = *(volatile type_name*)index_oop_from_field_offset_long(p, offset)


zzambers:
Log from problematic tests (without fix, for reference):
[CopyMemory-segfault.log](https://github.com/user-attachments/files/16415629/CopyMemory-segfault.log)
[OpaqueAccesses-segfault.log](https://github.com/user-attachments/files/16415630/OpaqueAccesses-segfault.log)

zzambers:
keep open

jerboaa:
@adinn @tstuefe @theRealAph Could you please help review this? We see crashes of this in GHA, so it looks to fix a real bug. Then again it's fairly late to touch this area in JDK 8u. Thoughts?

zzambers:
Btw, as GHA logs have expired and this is now last remaining issue showing in tier1 on Mac, I have tried to rebase these changes in separate branch (to retest on head).
There are [no failures except for security_infra](https://github.com/zzambers/jdk8u-dev/actions/runs/13524188699) tests - which are unrelated

(I can force-push rebased change here if desired)

hotspot/src/share/vm/prims/unsafe.cpp
```diff
@@ -167,11 +167,11 @@ jint Unsafe_invocation_key_to_method_slot(jint key) {
 
 #define GET_FIELD(obj, offset, type_name, v) \
   oop p = JNIHandles::resolve(obj); \
-  type_name v = *(type_name*)index_oop_from_field_offset_long(p, offset)
+  type_name v = *(volatile type_name*)index_oop_from_field_offset_long(p, offset)
```
adinn, line 170: This patch leaves it rather unclear what is being done here and why. A maintainer might easily be confused as to why `GET_FIELD` and `GET_FIELD_VOLATILE` both treat the field as having a volatile type. Likewise for SET_FIELD and SET_FIELD_VOLATILE.

One solution would be to add a comment here. However, I think it would be clearer to leave `GET_FIELD` and `SET_FIELD` unchanged here and to modify macro `DEFINE_GETSETOOP` so that it appends the volatile prefix to the type parameter passed in callsto `GET_FIELD` and `SET_FIELD` also adding a comment to explain why that is done. i.e.

```
// Note that the C type passed to GET_FIELD and SET_FIELD is prefixed with a
// volatile qualifier in order to ensure that the C compiler does not reorder
// the object field access wrt to preceding and succeeding volatile accesses
// to the thread flag field (made by UnsafeWrapper) which safeguard the object
// field access. See JDK-8186787.

#define DEFINE_GETSETOOP(jboolean, Boolean) \
 \
UNSAFE_ENTRY(jboolean, Unsafe_Get##Boolean##140(JNIEnv *env, jobject unsafe, jobject obj, jint offset)) \
  UnsafeWrapper("Unsafe_Get"#Boolean); \
  if (obj == NULL)  THROW_0(vmSymbols::java_lang_NullPointerException()); \
  GET_FIELD(obj, offset, volatile jboolean, v); \
  return v; \
UNSAFE_END \
 \
UNSAFE_ENTRY(void, Unsafe_Set##Boolean##140(JNIEnv *env, jobject unsafe, jobject obj, jint offset, jboolean x)) \
  UnsafeWrapper("Unsafe_Set"#Boolean); \
  if (obj == NULL)  THROW(vmSymbols::java_lang_NullPointerException()); \
  SET_FIELD(obj, offset, volatile jboolean, x); \
UNSAFE_END \
 \
UNSAFE_ENTRY(jboolean, Unsafe_Get##Boolean(JNIEnv *env, jobject unsafe, jobject obj, jlong offset)) \
  UnsafeWrapper("Unsafe_Get"#Boolean); \
  GET_FIELD(obj, offset, volatile jboolean, v); \
  return v; \
UNSAFE_END \
 \
UNSAFE_ENTRY(void, Unsafe_Set##Boolean(JNIEnv *env, jobject unsafe, jobject obj, jlong offset, jboolean x)) \
  UnsafeWrapper("Unsafe_Set"#Boolean); \
  SET_FIELD(obj, offset, volatile jboolean, x); \
UNSAFE_END \
 \
// END DEFINE_GETSETOOP.
```

adinn:
@jerboaa I think this is fine to backport. Adding a volatile qualifier to the access cannot really do any harm as it has the very limited effect of stopping the C++ compiler from reordering the volatile field access write relative to other volatile field accesses within the current thread's instruction stream. Since that is actually what is required here I cannot see any real risk. My real concern is that the introduction of volatile is clearly flagged in comments and happens at a point where it is most obvious what is going on and why.

hotspot/src/share/vm/prims/unsafe.cpp
adinn: @zzambers @jerboaa

Suggestion:

I'd really prefer it if we could rename the parameters `jboolean` and `Boolean` in macro `DEFINE_GETSETOOP` to something like `ctypename` and `javatypename`. That makes what is going on much clearer than with the current version -- where, somewhat bizarrely, the parameters are named using the arguments passed at the first call.

However, that's not a necessary change and the policy is not to change things that ain't broke. So, I'm not going to push this point. Whatever you feel is appropriate.

zzambers: @adinn I think passing `volatile` as part of `type_name` would be problematic, because it is also used in `SET_FIELD` macro to derive name of another macro `truncate_##type_name(x)`. Comment sound like a good idea.

adinn: Ah, yeah -- that's a shame but I cannot see a way round it.

Ok, so then just place the following comment above the definitions of `GET_FIELD` and `SET_FIELD`

```
// Note that GET_FIELD and SET_FIELD cast the object field's C type to
// volatile at the point of access. That is needed in order to ensure
// that the C compiler does not reorder the object field access wrt to
// preceding and succeeding volatile accesses to the thread flag field
// (made by UnsafeWrapper) which safeguard the field access. See
// JDK-8186787 for details.
```

zzambers:
@adinn thank you for the review

I added comment, as you have suggested. (As in the end, approach with DEFINE_GETSETOOP is not used, I kept it as it is.)
Should I mark you as co-author?
 
I rebased changes on current master, to avoid some unrelated, already fixed failures in GHA testing.

adinn:
@zzambers It's fine as it is.

zzambers:
/approval request Fix to unsafe to avoid tier1 failures (crashes) on Mac, small change, low risk, GHA tests OK

gnu-andrew (approved):
With the absence of [JDK-8150921](https://bugs.openjdk.org/browse/JDK-8150921) in 8u, this is really a backport in spirit than one having much relation to the 11u patch. I appreciate the effort taken here to achieve the same goal in 8u's codebase. I also agree with Andrew Dinn about the value of the comment for explaining this change. As this is handled differently in 11 and later, it's fine to apply that only here.

PR #600, opened 2024-10-31
8342822: jdk8u432-b06 does not compile on AIX

varada1110:
Use of llabs() for jlong resolves the build error on AIX.
Performed jtreg testing for hotspot/test on both aix-ppc64 and linux-ppc64le and no related failures observed 

JBS : [JDK-8342822](https://bugs.openjdk.org/browse/JDK-8342822)

---------
### Progress
- [ ] [JDK-8342822](https://bugs.openjdk.org/browse/JDK-8342822) needs maintainer approval

### Issue
 * [JDK-8342822](https://bugs.openjdk.org/browse/JDK-8342822): jdk8u432-b06 does not compile on AIX (**Bug** - P2 - Requested) ⚠️ Issue is already resolved. Consider making this a "backport pull request" by setting the PR title to `Backport <hash>` with the hash of the original commit. See [Backports](https://wiki.openjdk.org/display/SKARA/Backports).

### Reviewers
 * [Francisco Ferrari Bihurriet](https://openjdk.org/census#fferrari) (@franferrax - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/600/head:pull/600` \
`$ git checkout pull/600`

`$ git checkout pull/600` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/600/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 600`

`$ git pr show -t 600`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/600.diff">https://git.openjdk.org/jdk8u-dev/pull/600.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jdk8u-dev/pull/600#issuecomment-2449513572)
</details>

### DIFF
--- a/hotspot/src/share/vm/opto/superword.hpp
+++ b/hotspot/src/share/vm/opto/superword.hpp
@@ -588,7 +588,8 @@ class SWPointer VALUE_OBJ_CLASS_SPEC {
         _scale == q._scale   &&
         _invar == q._invar   &&
         _negate_invar == q._negate_invar) {
-      jlong difference = llabs(java_subtract((jlong)_offset, (jlong)q._offset));
+      jlong output = java_subtract((jlong)_offset, (jlong)q._offset);
+      jlong difference = output >= 0 ? output : -1ll * output;
       jlong max_diff = (jlong)1 << 31;
       if (difference >= max_diff) {
         return NotComparable;
--- a/hotspot/src/share/vm/opto/superword.hpp
+++ b/hotspot/src/share/vm/opto/superword.hpp
@@ -30,7 +30,10 @@
 #include "opto/phaseX.hpp"
 #include "opto/vectornode.hpp"
 #include "utilities/growableArray.hpp"
+
+#ifdef AIX
 #include "cstdlib"
+#endif
 
 //
 //                  S U P E R W O R D   T R A N S F O R M
--- a/hotspot/src/share/vm/opto/superword.hpp
+++ b/hotspot/src/share/vm/opto/superword.hpp
@@ -588,7 +588,7 @@ class SWPointer VALUE_OBJ_CLASS_SPEC {
         _scale == q._scale   &&
         _invar == q._invar   &&
         _negate_invar == q._negate_invar) {
-      jlong difference = labs(java_subtract((jlong)_offset, (jlong)q._offset));
+      jlong difference = llabs(java_subtract((jlong)_offset, (jlong)q._offset));
       jlong max_diff = (jlong)1 << 31;
       if (difference >= max_diff) {
         return NotComparable;
--- a/hotspot/src/share/vm/opto/superword.hpp
+++ b/hotspot/src/share/vm/opto/superword.hpp
@@ -30,6 +30,7 @@
 #include "opto/phaseX.hpp"
 #include "opto/vectornode.hpp"
 #include "utilities/growableArray.hpp"
+#include "cstdlib"
 
 //
 //                  S U P E R W O R D   T R A N S F O R M
@@ -588,8 +589,7 @@ class SWPointer VALUE_OBJ_CLASS_SPEC {
         _scale == q._scale   &&
         _invar == q._invar   &&
         _negate_invar == q._negate_invar) {
-      jlong output = java_subtract((jlong)_offset, (jlong)q._offset);
-      jlong difference = output >= 0 ? output : -1ll * output;
+      jlong difference = abs(java_subtract((jlong)_offset, (jlong)q._offset));
       jlong max_diff = (jlong)1 << 31;
       if (difference >= max_diff) {
         return NotComparable;
--- a/hotspot/src/share/vm/opto/superword.hpp
+++ b/hotspot/src/share/vm/opto/superword.hpp
@@ -588,7 +588,7 @@ class SWPointer VALUE_OBJ_CLASS_SPEC {
         _scale == q._scale   &&
         _invar == q._invar   &&
         _negate_invar == q._negate_invar) {
-      jlong difference = abs(java_subtract((jlong)_offset, (jlong)q._offset));
+      jlong difference = labs(java_subtract((jlong)_offset, (jlong)q._offset));
       jlong max_diff = (jlong)1 << 31;
       if (difference >= max_diff) {
         return NotComparable;


varada1110:
/approval 8342822 request build error observed on aix is resolved with use of llabs(). Build and  jtreg testing(hotspot/test) successful

jerboaa:
@varada1110 Please enable GHA testing on your fork. Thanks!

varada1110:
Hi @jerboaa, Could you please review the code changes.
Thanks,
Varada

jerboaa:
> Could you please review the code changes.

Some of the orginal reviewers should do that. @rwestrel @franferrax could you please take a look? Thank you!

franferrax (approved):
Hi, I'm not a reviewer, but it looks good to me.

We are using:

```cpp
long long int llabs (long long int n);
```

Which should be fine for 32-bit systems because on those `jlong` is generally defined as `long long` <sup>(1)</sup>.

It should also work fine in 64-bit systems because on those `jlong` is generally defined as `long` <sup>(1)</sup>, and, in the LP64 data model, `long long` has the same width as `long` <sup>[(2)](https://en.cppreference.com/w/cpp/language/types#Properties)</sup>.

----

(1)

https://github.com/openjdk/jdk8u-dev/blob/618917eb093243de2c5d7e83d4688bfe9ad04985/hotspot/src/cpu/ppc/vm/jni_ppc.h#L47-L51
https://github.com/openjdk/jdk8u-dev/blob/618917eb093243de2c5d7e83d4688bfe9ad04985/hotspot/src/cpu/aarch64/vm/jni_aarch64.h#L47-L51
https://github.com/openjdk/jdk8u-dev/blob/618917eb093243de2c5d7e83d4688bfe9ad04985/hotspot/src/cpu/zero/vm/jni_zero.h#L46-L50
https://github.com/openjdk/jdk8u-dev/blob/618917eb093243de2c5d7e83d4688bfe9ad04985/hotspot/src/cpu/x86/vm/jni_x86.h#L47-L51
https://github.com/openjdk/jdk8u-dev/blob/618917eb093243de2c5d7e83d4688bfe9ad04985/hotspot/src/cpu/sparc/vm/jni_sparc.h#L42-L46

franferrax:
@varada1110: the cause of this build error seems to be that while `jdk8u` uses the C++98 standard, the `long long` overload for `abs` was introduced in C++11:

```cpp
long long int abs (long long int n);
```

Haven't you found the same problem with `jdk11u`? At least in _Linux x64_ `src/hotspot/share/opto/superword.cpp` is built with `-std=gnu++98`.

varada1110:
Hi @franferrax ,
I couldn't find the use of abs in jdk11u in superword.cpp. For the safer side I have removed the use of abs and added the traditional way of doing the same. I have pushed the new code change and the testing for hotspot/jtreg is successful on aix-ppc didn't found any related failures. 
Thanks,
Varada

franferrax:
Hi @varada1110,

Here is the jdk11u version of the failing code: [src/hotspot/share/opto/superword.hpp:710](https://github.com/openjdk/jdk11u/blob/jdk-11.0.25-ga/src/hotspot/share/opto/superword.hpp#L710). We may experience the same build error on AIX.

I'm realizing my assumptions in the understanding of the issue were wrong. I thought it was a 32-bit build, but the build is for `aix-ppc64`. Thus, `jlong` should be defined as `long`, and the C++98 [`long int abs (long int n)` overload](https://en.cppreference.com/w/cpp/numeric/math/abs) (defined in `stdlib.h`, included from [globalDefinitions_xlc.hpp](https://github.com/openjdk/jdk8u-dev/blob/jdk8u432-ga/hotspot/src/share/vm/utilities/globalDefinitions_xlc.hpp#L40)) should work.

The problem seems to be in choosing between two equally suitable `abs()` definitions. A possible cause might be one definition coming from `stdlib.h`, the other one from `cstdlib` combined with a `using namespace std;` statement. Another less likely possibility is a compiler bug (see [IV24392](https://www.ibm.com/support/pages/apar/IV24392) and [LI71532](https://www.ibm.com/support/pages/apar/LI71532)).

@varada1110 / @andrew-m-leonard: could you help me understand the root cause of this build issue? Did the `xlC` compiler generate any informational `(I)` messages besides the `1540-0219 (S)` severe error?

With regard to the last change (8d18c5596af6db8c73fac2e072cdfe688dd36e83), I think it would be cleaner to just use the `ABS` template function:
https://github.com/openjdk/jdk8u-dev/blob/618917eb093243de2c5d7e83d4688bfe9ad04985/hotspot/src/share/vm/utilities/globalDefinitions.hpp#L1156

It's already being used in the file changed by this PR:
https://github.com/openjdk/jdk8u-dev/blob/618917eb093243de2c5d7e83d4688bfe9ad04985/hotspot/src/share/vm/opto/superword.hpp#L292-L295

But I think we'll first want to understand why the build failed, and if it isn't failing in jdk11u, what's causing the difference in behavior.

varada1110:
```
/opt/IBM/xlC/13.1.3/bin/xlC version.cpp

/opt/IBM/xlC/13.1.3/bin/.orig/xlC: 1501-216 (W) command option -64 is not recognized - passed to ld
"version.cpp", line 22.39: 1540-0219 (S) The call to "abs" has no best match.
"version.cpp", line 22.43: 1540-1229 (I) Argument number 1 is an rvalue of type "unsigned long".
"/opt/IBM/xlC/13.1.3/include/cstdlib", line 74.15: 1540-1202 (I) No candidate is better than "abs(long)".
"version.cpp", line 22.43: 1540-1231 (I) The conversion from argument number 1 to "long" uses "an integral conversion".
"/usr/include/stdlib.h", line 819.33: 1540-1202 (I) No candidate is better than "abs(int)".
"version.cpp", line 22.43: 1540-1231 (I) The conversion from argument number 1 to "int" uses "an integral conversion".
"version.cpp", line 22.44: 1540-0840 (W) The integer literal "43898989878" is out of range.

 /opt/IBM/xlC/13.1.3/bin/xlC -q64 version.cpp
abs(-43898989878) = 43898989878
```

I found that for long long int abs can overload with 64 bit compilation. The above error is observed while building jdk8 on AIX. 
Also with ABS template function we could resolve build error.

gnu-andrew:
Where are we with this change? It doesn't seem that it is ready for approval,  being still under review.
It sounds like we need to know why this is being considered 8u only, when the same `abs` call is used in later JDK versions.

andrew-m-leonard:
That's an interesting question, and I suspect that must be because jdk8u is
built with IBM XL C/C++ v13.1.3, whereas jdk11u v16.1.0.

franferrax:
Hi, sorry for my delayed reply, I'm resuming work on this issue.

Here is [the relevant part of the failure log](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-aix-ppc64-temurin/484/consoleText#:~:text=machnode.cpp-,%5B2024%2D11%2D14T21%3A10%3A24.721Z%5D,Error%201) (from [`jdk8u-aix-ppc64-temurin` Build #484](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-aix-ppc64-temurin/484/) thanks @jerboaa for pointing me to that Jenkins instance):

```
[2024-11-14T21:10:24.721Z] "/home/jenkins/workspace/build-scripts/jobs/jdk8u/jdk8u-aix-ppc64-temurin/workspace/build/src/hotspot/src/share/vm/opto/superword.hpp", line 591.26: 1540-0219 (S) The call to "abs" has no best match.
[2024-11-14T21:10:24.721Z] "/home/jenkins/workspace/build-scripts/jobs/jdk8u/jdk8u-aix-ppc64-temurin/workspace/build/src/hotspot/src/share/vm/opto/superword.hpp", line 591.43: 1540-1229 (I) Argument number 1 is an rvalue of type "long".
[2024-11-14T21:10:24.721Z] "/usr/include/math.h", line 1033.20: 1540-1202 (I) No candidate is better than "abs(long double)".
[2024-11-14T21:10:24.721Z] "/home/jenkins/workspace/build-scripts/jobs/jdk8u/jdk8u-aix-ppc64-temurin/workspace/build/src/hotspot/src/share/vm/opto/superword.hpp", line 591.43: 1540-1231 (I) The conversion from argument number 1 to "long double" uses "a floating point-integral conversion".
[2024-11-14T21:10:24.721Z] "/usr/include/math.h", line 1003.14: 1540-1202 (I) No candidate is better than "abs(float)".
[2024-11-14T21:10:24.721Z] "/home/jenkins/workspace/build-scripts/jobs/jdk8u/jdk8u-aix-ppc64-temurin/workspace/build/src/hotspot/src/share/vm/opto/superword.hpp", line 591.43: 1540-1231 (I) The conversion from argument number 1 to "float" uses "a floating point-integral conversion".
[2024-11-14T21:10:24.721Z] "/usr/include/math.h", line 1001.15: 1540-1202 (I) No candidate is better than "abs(double)".
[2024-11-14T21:10:24.721Z] "/home/jenkins/workspace/build-scripts/jobs/jdk8u/jdk8u-aix-ppc64-temurin/workspace/build/src/hotspot/src/share/vm/opto/superword.hpp", line 591.43: 1540-1231 (I) The conversion from argument number 1 to "double" uses "a floating point-integral conversion".
[2024-11-14T21:10:24.721Z] "/usr/include/stdlib.h", line 803.18: 1540-1202 (I) No candidate is better than "abs(int)".
[2024-11-14T21:10:24.721Z] "/home/jenkins/workspace/build-scripts/jobs/jdk8u/jdk8u-aix-ppc64-temurin/workspace/build/src/hotspot/src/share/vm/opto/superword.hpp", line 591.43: 1540-1231 (I) The conversion from argument number 1 to "int" uses "an integral conversion".
[2024-11-14T21:10:24.721Z] gmake[6]: *** [/home/jenkins/workspace/build-scripts/jobs/jdk8u/jdk8u-aix-ppc64-temurin/workspace/build/src/hotspot/make/aix/makefiles/rules.make:151: loopnode.o] Error 1
```

We are building for _ppc64_, and searching the best candidate for `abs(long)`. The following candidates are considered at an equal-level (so none of them is the "best match"):

* `abs(long double)` from `/usr/include/math.h`, using "a floating point-integral conversion"
* `abs(float)` from `/usr/include/math.h`, using "a floating point-integral conversion"
* `abs(double)` from `/usr/include/math.h`, using "a floating point-integral conversion"
* `abs(int)` from `/usr/include/stdlib.h`, using "an integral conversion"

`xlC` is unable to find the `abs(long)` candidate, which should qualify at the highest level (no argument conversion, so the "best match").

From @varada1110 [tests](#issuecomment-2487668534), this candidate is available in `cstdlib` (although in that particular case, isn't a "best match" for `unsigned long`):

```
"/opt/IBM/xlC/13.1.3/include/cstdlib", line 74.15: 1540-1202 (I) No candidate is better than "abs(long)".
```

Trying to get some external clues, I did a simple _Fedora Linux x86-64_ test with GCC. I found that `std::abs` is used even if we include `stdlib.h` and don't qualify `abs` with the `std::` namespace:

```c++
g++ -pedantic -Wall -xc++ -std=c++11 -op -<<'EOF'
//////////////////////////////////////////////////
#include <iostream>
#include <stdlib.h>

int main() {
    long var = -1234567890; // 0xffffffffb669fd2e
    std::cout << abs(var) << std::endl;
}
//////////////////////////////////////////////////
EOF
./p
sha256sum p
objdump --insn-width=8 -dC p | grep -A10 'abs\|<main>'
rm -f p
```

Output:

```asm
1234567890
3657bd6d3bb686f0dc8e058cce73199bb38b7c20113f3c599a11629577dbb12b  p
0000000000401146 <main>:
  401146:	55                      	push   %rbp
  401147:	48 89 e5                	mov    %rsp,%rbp
  40114a:	48 83 ec 10             	sub    $0x10,%rsp
  40114e:	48 c7 45 f8 2e fd 69 b6 	movq   $0xffffffffb669fd2e,-0x8(%rbp)
  401156:	48 8b 45 f8             	mov    -0x8(%rbp),%rax
  40115a:	48 89 c7                	mov    %rax,%rdi
  40115d:	e8 21 00 00 00          	call   401183 <std::abs(long)>
  401162:	48 89 c6                	mov    %rax,%rsi
  401165:	bf 40 40 40 00          	mov    $0x404040,%edi
  40116a:	e8 e1 fe ff ff          	call   401050 <std::ostream::operator<<(long)@plt>
  40116f:	be 30 10 40 00          	mov    $0x401030,%esi
  401174:	48 89 c7                	mov    %rax,%rdi
  401177:	e8 c4 fe ff ff          	call   401040 <std::ostream::operator<<(std::ostream& (*)(std::ostream&))@plt>
  40117c:	b8 00 00 00 00          	mov    $0x0,%eax
  401181:	c9                      	leave
  401182:	c3                      	ret

0000000000401183 <std::abs(long)>:
  401183:	55                      	push   %rbp
  401184:	48 89 e5                	mov    %rsp,%rbp
  401187:	48 89 7d f8             	mov    %rdi,-0x8(%rbp)
  40118b:	48 8b 45 f8             	mov    -0x8(%rbp),%rax
  40118f:	48 89 c2                	mov    %rax,%rdx
  401192:	48 f7 da                	neg    %rdx
  401195:	48 0f 49 c2             	cmovns %rdx,%rax
  401199:	5d                      	pop    %rbp
  40119a:	c3                      	ret
```

In fact, the following three combinations produce an identical binary:

* `#include <stdlib.h>` / `std::abs(var)`
* `#include <stdlib.h>` / `abs(var)` (the combination in the previous snippet)
* `#include <cstdlib>` / `std::abs(var)`

Apparently, the `#include <cstdlib>` / `abs(var)` combination yields some sort of inlined `abs` (I don't know where it's coming from):

```asm
0000000000401146 <main>:
  [...]
  40114e:	48 c7 45 f8 2e fd 69 b6 	movq   $0xffffffffb669fd2e,-0x8(%rbp)
  401156:	48 8b 45 f8             	mov    -0x8(%rbp),%rax
  40115a:	89 c2                   	mov    %eax,%edx
  40115c:	f7 da                   	neg    %edx
  40115e:	0f 49 c2                	cmovns %edx,%eax
  [...]
```

I will proceed to get [Temporary test system access](https://github.com/adoptium/infrastructure/blob/master/docs/AccessControl.md#temporary-test-system-access) and continue experimentation in _AIX ppc64_.

Some doubts I plan to clarify are:

1. What are the _AIX ppc64_ results of the equivalent tests I did for _Fedora Linux x86-64_?
2. Assuming I get access to both compiler versions, are there any differences between `xlC` 13.1.3 and `xlC` 16.1.0?
3. Is there a `/opt/IBM/xlC/13.1.3/include/stdlib.h` file?
    * What happens if we include that one instead of `/usr/include/stdlib.h`?
    * Assuming I get access to both compiler versions, repeat for `/opt/IBM/xlC/16.1.0/include/stdlib.h`

Any other research suggestions or advices are welcome.

varada1110:
The build error persists with xlc 16.1.0 version. Also observed that with <iostream.h> or adding inline abs() function resolves the issue. https://github.com/openjdk/jdk8u-dev/compare/master...varada1110:abs?expand=1
```
--- a/hotspot/src/share/vm/libadt/port.hpp
+++ b/hotspot/src/share/vm/libadt/port.hpp
@@ -91,7 +91,7 @@ inline int max( int a, int b) { return a > b ? a : b; }
 #define IMPLEMENTATION
 #include <stdlib.h>
 #include <memory.h>
-
+inline long abs( long x ) { return x < 0 ? -x : x; }
+//#include <iostream.h>
 #elif defined(_MSC_VER)
 // Microsoft Visual C++
 //#define INTERFACE
```

franferrax:
Hi, after getting access to an _AIX system_ (thanks @sxa) I have a clearer understanding of what is going on.

# Tests I executed in _AIX 7.2_

All the tests have the following structure, I will specify the C++ snippet and the results.

```bash
cat >test.cpp <<'EOF'
// C++ snippet goes here
EOF
/opt/IBM/xlC/13.1.3/bin/xlC_r -q64 -c -o/dev/null test.cpp
/opt/IBM/xlC/16.1.0/bin/xlC_r -q64 -c -o/dev/null test.cpp
rm -f test.cpp
```

## Finding the combination that causes the failure

### Snippets that worked

All the following snippets worked, both with `xlC` 13.1.3 and 16.1.0.

#### Use `stdlib.h` alone, with `abs()`

```cpp
#include <stdlib.h>

long test() {
    return abs((long)1);
}
```

#### Use `cstdlib` alone, with `std::abs()`

```cpp
#include <cstdlib>

long test() {
    return std::abs((long)1);
}
```

#### Use `cstdlib` with `std::abs()`, but also include `cmath` before `cstdlib`

```cpp
#include <cmath>
#include <cstdlib>

long test() {
    return std::abs((long)1);
}
```

#### Use `cstdlib` with `std::abs()`, but also include `cmath` after `cstdlib`

```cpp
#include <cstdlib>
#include <cmath>

long test() {
    return std::abs((long)1);
}
```

### Snippets that failed

The following snippets failed in the same way, both with `xlC` 13.1.3 and 16.1.0. See the error at the end.

#### Use `stdlib.h` with `abs()`, but also include `cmath` before `cstdlib`

```cpp
#include <cmath>
#include <stdlib.h>

long test() {
    return abs((long)1);
}
```

#### Use `stdlib.h` with `abs()`, but also include `cmath` after `cstdlib`

```cpp
#include <stdlib.h>
#include <cmath>

long test() {
    return abs((long)1);
}
```

#### Error

```
"test.cpp", line 5.12: 1540-0219 (S) The call to "abs" has no best match.
"test.cpp", line 5.16: 1540-1229 (I) Argument number 1 is an rvalue of type "long".
"/usr/include/math.h", line 1033.20: 1540-1202 (I) No candidate is better than "abs(long double)".
"test.cpp", line 5.16: 1540-1231 (I) The conversion from argument number 1 to "long double" uses "a floating point-integral conversion".
"/usr/include/math.h", line 1003.14: 1540-1202 (I) No candidate is better than "abs(float)".
"test.cpp", line 5.16: 1540-1231 (I) The conversion from argument number 1 to "float" uses "a floating point-integral conversion".
"/usr/include/math.h", line 1001.15: 1540-1202 (I) No candidate is better than "abs(double)".
"test.cpp", line 5.16: 1540-1231 (I) The conversion from argument number 1 to "double" uses "a floating point-integral conversion".
"/usr/include/stdlib.h", line 829.18: 1540-1202 (I) No candidate is better than "abs(int)".
"test.cpp", line 5.16: 1540-1231 (I) The conversion from argument number 1 to "int" uses "an integral conversion".
```

### Snippets that resolve the issue

It seems the problem is the conjunction of `cmath` with `stdlib.h`, this looks pretty similar to [IY68511: NO ABS(LONG) IMPLEMENTATION IN \<CSTDLIB\>](https://www.ibm.com/support/pages/apar/IY68511):
> ## Error description
>
> There is currently no implementation for `abs(long)` functions in `<cstdlib>`, but `abs(long)` is defined in the C++ Standard.
>
> ## Local fix
>
> Use `<cmath>` without `<cstdlib>`
>
> ## Problem summary
>
> `abs(long)` was not defined in the library implementation.
>
> ## Problem conclusion
>
> In order to use `abs(long)`, the macro `__ABS_LONG__` must be defined if using `stdlib.h` instead of `cstdlib`, or if `stdlib.h` occurs before `cstdlib` in the include path.

The following snippets make it work by adding a third element (inspiration from *Problem conclusion*).

#### Define the `__ABS_LONG__` macro as suggested

```cpp
#define __ABS_LONG__
#include <stdlib.h>
#include <cmath>

long test() {
    return abs((long)1);
}
```

NOTE: `__ABS_LONG__` must be defined at the beginning, any combination defining `__ABS_LONG__` after some of the includes, fails.

#### Include `cstdlib` in addition to `stdlib.h`

```cpp
#include <cstdlib>
#include <stdlib.h>
#include <cmath>

long test() {
    return abs((long)1);
}
```

NOTES:

* This succeeds for all the 6 possible ways of arranging the 3 includes
* If I replace `cstdlib` with `iostream` I get the same results (for the 6 possible arrangements)
    * I don't see a direct `#include <cstdlib>` from `iostream`, but I assume an indirect inclusion or this somehow implying the definition of `__ABS_LONG__`
    * This originally made it more difficult to reproduce the issue, as I was basing on my previous snippets, which were including `iostream` to use `std::cout`

# Conclusions

For 11u, I would discard the compiler version as a possible reason why we aren't experiencing this issue (given all the previous tests were identical in both versions). I would lean towards a different preprocessor result, due to a different topology of the includes graph.

For 8u, in my opinion, the acceptable fixes would include:

* Passing `-D__ABS_LONG__` to `xlC`
* Including `cstdlib` in `hotspot/src/share/vm/opto/superword.hpp`, which is the one using `abs(long)`
    * This shouldn't break other platforms builds, but we would need to test them

I've tested both by re-executing the failed build command (not the whole build) and seem to be working. Here is the build command I've obtained building with `make_64 LOG=debug CONF=aix-ppc64-normal-server-release images`, in case anyone wants to give a look to the compiler flags:

```bash
# PWD: /home/fferrari/jdk8u/build/aix-ppc64-normal-server-release/hotspot/aix_ppc64_compiler2/product
rm -f loopnode.o
/opt/IBM/xlC/13.1.3/bin/xlC_r -DAIX -DPPC64 -DPRODUCT \
  -I/home/fferrari/jdk8u/hotspot/src/share/vm/prims \
  -I/home/fferrari/jdk8u/hotspot/src/share/vm \
  -I/home/fferrari/jdk8u/hotspot/src/share/vm/precompiled \
  -I/home/fferrari/jdk8u/hotspot/src/cpu/ppc/vm \
  -I/home/fferrari/jdk8u/hotspot/src/os_cpu/aix_ppc/vm \
  -I/home/fferrari/jdk8u/hotspot/src/os/aix/vm \
  -I/home/fferrari/jdk8u/hotspot/src/os/posix/vm \
  -I../generated \
  -DHOTSPOT_RELEASE_VERSION="\"25.442-b00\"" \
  -DHOTSPOT_BUILD_TARGET="\"product\"" -DHOTSPOT_BUILD_USER="\"fferrari\"" \
  -DHOTSPOT_LIB_ARCH=\"ppc64\" -DHOTSPOT_VM_DISTRO="\"OpenJDK\"" \
  -DTARGET_OS_FAMILY_aix -DTARGET_ARCH_ppc -DTARGET_ARCH_MODEL_ppc_64 \
  -DTARGET_OS_ARCH_aix_ppc -DTARGET_OS_ARCH_MODEL_aix_ppc_64 \
  -DTARGET_COMPILER_xlc -DINCLUDE_JFR=0 -DCOMPILER2 -qpic=large -qnortti \
  -qnoeh -D_REENTRANT -D__STDC_FORMAT_MACROS -q64 -O3 -qalias=noansi -g \
  -qarch=ppc64 -qtune=balanced -qinlglue -qstrict -qxflag=ASMMIDCOALFIX \
  -qxflag=asmfastsync -qhot=level=1 -qignerrno -qinline -q64 -qlanglvl=c99vla \
  -qlanglvl=noredefmac -qsuppress=1540-0198 -qsuppress=1540-1090 \
  -qsuppress=1500-010 -qsuppress=1540-1639 -qsuppress=1540-1088 \
  -lpthread -DALLOW_OPERATOR_NEW_USAGE -c -qmakedep=gcc \
  -MF ../generated/dependencies/loopnode.o.d -o loopnode.o \
  /home/fferrari/jdk8u/hotspot/src/share/vm/opto/loopnode.cpp
```

varada1110:
Thank you, @franferrax, for identifying the root cause of the issue.
 I have added the <cstdlib> header file, which is resolving the build error. Additionally, I checked by adding -D__ABS_LONG__ in the AIX configuration, and that also resolves the issue. Testing for hotspot/test is successful, with no related failures observed.

sxa:
> Thank you, @franferrax, for identifying the root cause of the issue. I have added the header file, which is resolving the build error. Additionally, I checked by adding -D__ABS_LONG__ in the AIX configuration, and that also resolves the issue. Testing for hotspot/test is successful, with no related failures observed.

This seems to break Solaris which can't find the new header file:
`09:40:41  "/export/home/jenkins/workspace/build-scripts/jobs/jdk8u/jdk8u-solaris-x64-temurin/workspace/build/src/hotspot/src/share/vm/opto/superword.hpp", line 33: Error: Could not open include file "cstdlib".`

Ref: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-x64-temurin/524/consoleText

varada1110:
I have enclosed the header file within an #ifdef and made it specific to AIX to ensure compatibility

franferrax (approved):
Hi @varada1110,

Great! The change looks good to me, is based on the root cause, and should be 100% innocuous for other platforms.

I would also like to know if this looks good to @rwestrel or @gnu-andrew. I'm not a reviewer and we will need approval from one anyway.

gnu-andrew:
> Hi @varada1110,
> 
> Great! The change looks good to me, is based on the root cause, and should be 100% innocuous for other platforms.
> 
> I would also like to know if this looks good to @rwestrel or @gnu-andrew. I'm not a reviewer and we will need approval from one anyway.

It looks pretty risk-free to me, if it is only applied where it is needed (on AIX)

I would suggest moving this PR to github.com/openjdk/jdk8u as a regression fix aimed at the January release. Otherwise, 8u442 will be broken on AIX as well (8u-dev is now 8u452, due for release in April 2025).

Thanks for the detailed analysis and testing. I wasn't aware there were public build runs on AIX either. Maybe this is worth referencing on https://wiki.openjdk.org/display/Build/Supported+Build+Platforms ?

franferrax:
# Bonus

> For 11u, I would discard the compiler version as a possible reason why we aren't experiencing this issue (given all the previous tests were identical in both versions). I would lean towards a different preprocessor result, due to a different topology of the includes graph.

I was wrong, after a quick trick to ignore a missing `fontconfig` dependency, I was able to start a 11u build (I'm not interested in finishing it). By executing with `make_64 LOG=debug CONF=aix-ppc64-normal-server-release images`, I was able to retrieve the build command, and discovered that the compiler employed is `/opt/IBM/xlC/16.1.0/bin/xlclang++` (not `/opt/IBM/xlC/16.1.0/bin/xlC_r`).

So the previous tests were wrong, and in fact `xlclang++` 16.1.0 is not affected by the issue, the following snippet builds without problems:

```cpp
cat >test.cpp <<'EOF'
#include <cmath>
#include <stdlib.h>

long test() {
    return abs((long)1);
}
EOF
/opt/IBM/xlC/16.1.0/bin/xlclang++ -q64 -c -o/dev/null test.cpp
rm -f test.cpp
```

So we won't hit this in 11u. For reference, here is the full `loopnode.o` build command:

```bash
/opt/IBM/xlC/16.1.0/bin/xlclang++ -qmakedep=gcc \
  -MF ./build/aix-ppc64-normal-server-release/hotspot/variant-server/libjvm/objs/loopnode.d.tmp \
  -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -D__STDC_CONSTANT_MACROS -D_REENTRANT \
  -qtbtable=full -qtune=balanced -qalias=noansi -qstrict -qtls=default -qnortti -qnoeh \
  -qignerrno -qstackprotect -DLIBC=default -DSUPPORTS_CLOCK_MONOTONIC -DAIX \
  -qpic=large -qarch=ppc64 -q64 -DNDEBUG -DPRODUCT -DTARGET_ARCH_ppc -DINCLUDE_SUFFIX_OS=_aix \
  -DINCLUDE_SUFFIX_CPU=_ppc -DINCLUDE_SUFFIX_COMPILER=_xlc -DTARGET_COMPILER_xlc -DPPC64 \
  -DHOTSPOT_LIB_ARCH='"ppc64"' -DCOMPILER1 -DCOMPILER2 -DINCLUDE_JVMCI=0 -DINCLUDE_CDS=0 \
  -DINCLUDE_AOT=0 -DINCLUDE_ZGC=0 -DINCLUDE_SHENANDOAHGC=0 -DINCLUDE_JFR=0 \
  -I ./build/aix-ppc64-normal-server-release/hotspot/variant-server/gensrc/adfiles \
  -I ./src/hotspot/share \
  -I ./src/hotspot/os/aix \
  -I ./src/hotspot/os/posix \
  -I ./src/hotspot/cpu/ppc \
  -I ./src/hotspot/os_cpu/aix_ppc \
  -I ./build/aix-ppc64-normal-server-release/hotspot/variant-server/gensrc \
  -I ./src/hotspot/share/precompiled \
  -I ./src/hotspot/share/include \
  -I ./src/hotspot/os/posix/include \
  -I ./build/aix-ppc64-normal-server-release/support/modules_include/java.base \
  -I ./build/aix-ppc64-normal-server-release/support/modules_include/java.base/aix \
  -I ./src/java.base/share/native/libjimage -q64 -DDONT_USE_PRECOMPILED_HEADER -g1 \
  -qsuppress=1540-0216 -qsuppress=1540-0198 -qsuppress=1540-1090 -qsuppress=1540-1639 \
  -qsuppress=1540-1088 -qsuppress=1500-010 -O3 -qhot=level=1 -qinline -qinlglue -c \
  -o ./build/aix-ppc64-normal-server-release/hotspot/variant-server/libjvm/objs/loopnode.o \
  ./src/hotspot/share/opto/loopnode.cpp
```

gnu-andrew:
> > I would suggest moving this PR to github.com/openjdk/jdk8u as a regression fix aimed at the January release. Otherwise, 8u442 will be broken on AIX as well (8u-dev is now 8u452, due for release in April 2025).
> 
> Agreed - that's what I was hoping for - that this would get accepted in time to fix 8u442 next month. @varada1110 Can you handle migrating the PR into jdk8u?

We need to do this ASAP as we freeze at the end of next week. I doubt many people will be around next week either, so it may already be too late.  I'll try and keep an eye out, but today is my last work day of 2024.

> 
> > Thanks for the detailed analysis and testing. I wasn't aware there were public build runs on AIX either. Maybe this is worth referencing on https://wiki.openjdk.org/display/Build/Supported+Build+Platforms ?
> 
> Yeah I'd be good with that. the post at https://api.adoptium.net/v3/binary/latest/21/ea/linux/aarch64/jdk/hotspot/normal/adoptium talks about the builds we publish for each EA tag that comes out. [The build link I referenced above](https://github.com/openjdk/jdk8u-dev/pull/600#issuecomment-2551256620) was a special that I ran specifically against that branch (That isn't done automatically!) but in general they can be pulled from the API links referenced in that article e.g. https://api.adoptium.net/v3/binary/latest/8/ea/aix/ppc64/jdk/hotspot/normal/adoptium for the latest AIX jdk8u EA tag (unless it failed to build of course, like it does for AIX, so that link is currently serving up jdk8u432-b05 as the last one that built successfully)

Is the first link wrong? It take me to an AArch64 binary, not a post. It is really good to know this is being regularly tested.

varada1110:
I was away and did not have access to my laptop. I have created the PR on jdk8u. Please review the code change, jtreg testing looks fine too
https://github.com/openjdk/jdk8u/pull/65

PR #634, opened 2025-03-07
8351439: [8u] test/java/util/TimeZone/tools/share/Makefile use wrong path to tzdb

mrserb:
The TZDATA has been updated to work with jdk8u-dev

### DIFF
--- a/jdk/test/java/util/TimeZone/tools/share/Makefile
+++ b/jdk/test/java/util/TimeZone/tools/share/Makefile
@@ -30,7 +30,7 @@
 #     make install
 #
 
-TZDATA = ../../../../../../../src/java.base/share/data/tzdata
+TZDATA = ../../../../../../make/data/tzdata/
 TZDATA_VER = `grep '^tzdata' $(TZDATA)/VERSION`
 TZNAME = africa antarctica asia australasia europe northamerica \
 	solar87 solar88 solar89 southamerica \


dimitryc (approved):

jerboaa (commented):
There should be no trailing slash, I think.

jerboaa (approved):

mrserb:
/approval request. Fix for a typo to simplify updating the TimeZone test during the tzdata integration process.

gnu-andrew:
What makes you think this is 8u only? The path wasn't changed until [JDK-8257733](https://bugs.openjdk.org/browse/JDK-8257733) in OpenJDK 19 and the file in question was backported from [JDK-8028265](https://bugs.openjdk.org/browse/JDK-8028265) in OpenJDK 20.

Do you actually make use of this `Makefile`? Given we usually backport the generated test files in question from trunk where the `java.base` path is correct, I can see why this has been missed for the past two and a half years.

mrserb:
>Do you actually make use of this Makefile? Given we usually backport the generated test files in question from trunk where the java.base path is correct, I can see why this has been missed for the past two and a half years.

Yes, you are right - starting with JDK 11 most of the time it's just about applying changes from trunk, so this file does not play a significant role. However, JDK 8 uses a different format, which requires manual handling and as part of that process we also regenerate the test data(just in case), which has helped uncover some missed changes recently.(this bug was fixed in our src looong time ago).

PR #639, opened 2025-03-22
8339810: Clean up the code in sun.tools.jar.Main to properly close resources and use ZipFile during extract

martinuy:
Hi,

I'd like to have a review for the 8u backport of [8339810](https://bugs.openjdk.org/browse/JDK-8339810).

Changes to the 11u patch:

 * Paths.

 * jdk/src/share/classes/sun/tools/jar/Main.java
   * Copyright date. Manually fixed.
   * 8u does not have 8199871, 8058520, 8114827, 8156497, 8142968, 8172432, 8217375, 8169069 so the context is different and changes do not apply cleanly.
   * I decided to include the part of 8217375 affecting Main.java because it's related to resources leakage, relevant to 8339810. Given that this part is an insignificant part of 8217375, I propose not add a reference and give the (misleading) impression that it's a full backport of it.

Testing:

 * No regressions found in jdk/sun/security/tools/jarsigner.

Thanks,
Martin.-

---------
### Progress
- [x] [JDK-8339810](https://bugs.openjdk.org/browse/JDK-8339810) needs maintainer approval

### Issue
 * [JDK-8339810](https://bugs.openjdk.org/browse/JDK-8339810): Clean up the code in sun.tools.jar.Main to properly close resources and use ZipFile during extract (**Bug** - P4 - Approved)

### Reviewers
 * [Yuri Nesterenko](https://openjdk.org/census#yan) (@yan-too - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/639/head:pull/639` \
`$ git checkout pull/639`

`$ git checkout pull/639` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/639/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 639`

`$ git pr show -t 639`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/639.diff">https://git.openjdk.org/jdk8u-dev/pull/639.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jdk8u-dev/pull/639#issuecomment-2745025254)
</details>

### DIFF
--- a/jdk/src/share/classes/sun/tools/jar/Main.java
+++ b/jdk/src/share/classes/sun/tools/jar/Main.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1996, 2013, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1996, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -150,7 +150,7 @@ class Main {
     /**
      * Starts main program with the specified arguments.
      */
-    public synchronized boolean run(String args[]) {
+    public synchronized boolean run(String[] args) {
         ok = true;
         if (!parseArgs(args)) {
             return false;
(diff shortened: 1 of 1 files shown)


jankratochvil:
It would be a little cleaner backport (preserving jdk8 code) with this change: 
[entries.patch.txt](https://github.com/user-attachments/files/19428804/entries.patch.txt)

martinuy:
> It would be a little cleaner backport (preserving jdk8 code) with this change: [entries.patch.txt](https://github.com/user-attachments/files/19428804/entries.patch.txt)

I couldn't track down the suggested change to any bug. Can you help me with that? I've seen that 8339810 for example keeps the comment `// add the remaining new files`.

jankratochvil:
> I couldn't track down the suggested change to any bug. Can you help me with that? I've seen that 8339810 for example keeps the comment `// add the remaining new files`.

It depends what is the goal. IMO during backporting https://github.com/openjdk/jdk/commit/8fce5275fc94ebc404a6a37f5ea0407140de63c1 does only reindent the block. So one should get the original jdk8 block and reindent it.

But then the comment really should be there, that is a mistake in my patch. The only difference is `(File f: entries)` vs. `(File f : entries)` which does not really matter. Let's forget about my comment.

martinuy:
Thanks for clarifying, @jankratochvil . Yes, I'd suggest to keep the code as it looks after the fix in newer releases (when possible) in the hope that the next overlapping backport applies cleanly.

yan-too (approved):

martinuy:
/approval request 8u is affected by this bug. I'd like to request a backport approval. Backport was reviewed.

gnu-andrew (approved):
All the later changes do make this quite tricky to review (and to backport, I imagine). After comparing the 11u & 8u patches (both as a diff and line by line) and the resulting patched files, I think this looks good.

The inclusion of the hunk from 8217375 seems right. Given that hunk is changed by this patch, it would be odd to not include that and instead alter the incoming hunk to be knowingly inferior to what is in 11u. To that end, I think it may be worth following up this change with an 8u specific change to bring over the similar changes to `run`, `create` and `copy` which are hiding in [8158295](def15478ebc213eeff8eb4da9178f8bac4c72604) and [8172432](https://github.com/openjdk/jdk11u/commit/cb9f76175c61478fcd24e340a943a8c0c6103202)

PR #635, opened 2025-03-10
8309841: Jarsigner should print a warning if an entry is removed

wkia:
This is backport of "8309841: Jarsigner should print a warning if an entry is removed"

Original patch does not apply cleanly to jdk8:
- some minor conflicts in JDK code
- files moved to appropriate locations
- added the check in `jdk/src/share/classes/sun/security/tools/jarsigner/Main.java` at line 1196

in tests:
- `ed25519` algorithm was replaced with `RSA` in `RemovedFiles.java`
- `JarEntry copyEntry()` procedure was manually added to `JarUtils.java` to make `JarUtilsTest` working.
- the tests were adapted to Java 8, including replaced `Path.of`, `Files.writeString`, package names, and arrays processing as well.

We need this fix in jdk8, as all versions have this issue with jarsigner.

New tests successfully ran locally on Linux, x86_64.

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change requires CSR request [JDK-8334261](https://bugs.openjdk.org/browse/JDK-8334261) to be approved
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] [JDK-8309841](https://bugs.openjdk.org/browse/JDK-8309841) needs maintainer approval

### Issues
 * [JDK-8309841](https://bugs.openjdk.org/browse/JDK-8309841): Jarsigner should print a warning if an entry is removed (**Enhancement** - P3 - Approved)
 * [JDK-8334261](https://bugs.openjdk.org/browse/JDK-8334261): Jarsigner should print a warning if an entry is removed (**CSR**)

### Reviewers
 * [Andrew John Hughes](https://openjdk.org/census#andrew) (@gnu-andrew - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/635/head:pull/635` \
`$ git checkout pull/635`

Update a local copy of the PR: \
`$ git checkout pull/635` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/635/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 635`

View PR using the GUI difftool: \
`$ git pr show -t 635`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
[...]

PR #640, opened 2025-03-24
Backport 2a791467919c9df9869e6fe1e57df0a5caa90d8f

wkia:
This is backport of "8240235: jdk.test.lib.util.JarUtils updates jar files incorrectly", needed for backporting "8309841: Jarsigner should print a warning if an entry is removed"

Original patch does not apply cleanly to jdk8, fixed minor conflict as `public static void updateJarFile(Path jarfile, Path dir, Path... files)` doesn't exist in jdk8.


gnu-andrew (approved):
Backport looks good. Fine with skipping the change to the missing method as discussed on [the backport of 8309841](https://github.com/openjdk/jdk8u-dev/pull/635#issuecomment-2743975905). GHA build and testing passes without regression.

Please apply for approval.

PR #641, opened 2025-03-26
8352942: jdk/jfr/startupargs/TestMemoryOptions.java fails with 32-bit build

tkiriyama:
Hi All,
I would like to add this bug fix for the bug in jdk/jfr/startupargs/TestMemoryOptions.java. This test contains 24 test cases and fails the "ThreadBufferSizeExceedMemorySize" case.
The cause of this bug is the memory allocation issue, which occurs only on 32-bit Server VM, not on Client VM or 64-bit JDK. The failure happens because Server VM's default heap size reduces available memory space, causing JFR to fail memory allocation.
To resolve this issue, -Xmx256M is explicitly set, matching the Client VM default heap size, ensuring sufficient memory space remains available for JFR.
I believe that this test verifies that the combination of memory options for JFR is valid or invalid and that the MaxHeapSize setting does not affect the verification.
Change has been verified locally, test-fix only, no risk.

Would someone please review this fix?

---------
### Progress
- [ ] [JDK-8352942](https://bugs.openjdk.org/browse/JDK-8352942) needs maintainer approval

### Issue
 * [JDK-8352942](https://bugs.openjdk.org/browse/JDK-8352942): jdk/jfr/startupargs/TestMemoryOptions.java fails with 32-bit build (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/641/head:pull/641` \
`$ git checkout pull/641`

`$ git checkout pull/641` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/641/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 641`

`$ git pr show -t 641`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/641.diff">https://git.openjdk.org/jdk8u-dev/pull/641.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jdk8u-dev/pull/641#issuecomment-2753649580)
</details>

### DIFF
--- a/jdk/test/jdk/jfr/startupargs/TestMemoryOptions.java
+++ b/jdk/test/jdk/jfr/startupargs/TestMemoryOptions.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -493,6 +493,7 @@ public class TestMemoryOptions {
             ProcessBuilder pb;
             if (flightRecorderOptions != null) {
                 pb = ProcessTools.createJavaProcessBuilder(true,
+                                                           "-Xmx256m",
                                                            flightRecorderOptions,
                                                            "-XX:StartFlightRecording",
                                                            SUT.class.getName(),
@@ -500,6 +501,7 @@ public class TestMemoryOptions {
             } else {
                 // default, no FlightRecorderOptions passed
                 pb = ProcessTools.createJavaProcessBuilder(true,
+                                                           "-Xmx256m",
                                                            "-XX:StartFlightRecording",
                                                            SUT.class.getName(),
                                                            tc.getTestName());


tkiriyama:
@phohensee 
Thank you for your comment.
I see. I'll check it. Sorry, let me confirm that first. Is 32-bit support being maintained in the later JDK versions (11, 17, 21, 24, tip)? If it is, I'm happy to fix this in the later JDKs.
It seems like there's already a lack of enthusiasm for 32-bit development in JDK 21.
https://openjdk.org/jeps/449

phohensee:
Yes, there is lack of enthusiasm for 32-bit ports, and the x86 32-bit port is in the process of being removed in JDK 25. But, the arm-32 port still exists in tip, and the LTS 32-bit ports are still supported.

tkiriyama:
@phohensee 

I am sorry for the late reply as it took time to verify.
I didn't have an arm environment, so I built a slightly older revision in an x86 environment.
As a result, with the default Xmx, the latest java hardly reproduced this problem.
If larger Xmx is used, the combination with the JFR option may cause an error.


```
$ jdk_32/jdk/bin/java -Xmx3221225472 -XX:FlightRecorderOptions=threadbuffersize=30m -XX:StartFlightRecording -version
[0.144s][warning][jfr,system] Memory allocation failed for size [31457304] bytes
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (malloc) failed to allocate 31457304 bytes. Error detail: AllocateHeap
# An error report file with more information is saved as:
# /work/hs_err_pid411102.log
[thread 411103 also had an error]

[error occurred during error reporting (), id 0xe0000001, Out of Memory Error (src/hotspot/share/jfr/utilities/jfrAllocation.cpp:84)]
```

It is unlikely that the tests will fail with the latest java, but the tests are potentially risky and I will fix it in the latest java.
This pull request is closed once.

PR #663, opened 2025-06-30
8360166: CodeSource.implies(): Wildcard host fails to imply specific host

Korov:
Please review this PR which use hostname not cname to check imply. And alse add test case.

If there are any problems please let me know and I will correct them promptly.


jerboaa:
This needs to go into later releases before a backport to JDK 8 can be evaluated.

PR #666, opened 2025-07-07
Backport 6e48caf2507d4a45e1861ee699a32a5e459d70c2

sendaoYan:
Hi all,
  This is clean backport to fix the test run time outed which the tested machine has huge cpu core number. Change has been verified locally, test-fix only, no risk.


sendaoYan:
/approval request Clean backport to fix the test run time outed which the tested machine has huge cpu core number. Change has been verified locally, test-fix only, no risk.

PR #685, opened 2025-08-29
8353657: [8u] Test tools/launcher/VersionCheck.java fails with debug build

sendaoYan:
Hi all,

Test tools/launcher/VersionCheck.java fails with fastdebug build which report 'jstack.debuginfo: cannot execute binary file'. This failure fixed by [JDK-8237192](https://bugs.openjdk.org/browse/JDK-8237192) in jdk15, only a subset of [JDK-8237192](https://bugs.openjdk.org/browse/JDK-8237192) to fix this test failure. So it's not suitable to backport [JDK-8237192](https://bugs.openjdk.org/browse/JDK-8237192) to jdk8u-dev, then I create this PR to fix this testbug.

This PR skip all the `*.debuginfo` files to check the '-version' or '-J-version'. Change has been verified localy, test-fix only, no risk.

### DIFF
--- a/jdk/test/tools/launcher/TestHelper.java
+++ b/jdk/test/tools/launcher/TestHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2013, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2025, Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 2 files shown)


sendaoYan:
/approval request Fix the test bug which will cause test fail with jdk fastdebug build. Change has been verified locally, test-fix only, no risk.

PR #689, opened 2025-09-03
Backport 4c9eaddaef83c6ba30e27ae3e0d16caeeec206cb

johnshajiang:
Please review this backport of `JDK-8364597: Replace THL A29 Limited with Tencent`.
This backport is not clean, because most files don't exist in jdk8u,
and the copyright years in the following files are different between the repos.
- jdk/test/jdk/internal/platform/docker/GetFreeSwapSpaceSize.java
- jdk/test/jdk/internal/platform/docker/TestGetFreeSwapSpaceSize.java

### DIFF
--- a/hotspot/test/compiler/unsafe/TestMisalignedUnsafeAccess.java
+++ b/hotspot/test/compiler/unsafe/TestMisalignedUnsafeAccess.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020 THL A29 Limited, a Tencent company. All rights reserved.
+ * Copyright (C) 2020, Tencent. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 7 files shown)


johnshajiang:
/approval request for backport of JDK-8364597: Replace THL A29 Limited with Tencent

Because THL A29 Limited has been dissolved, so it has to update the copyright notes accordingly.

gnu-andrew:
It looks like you are not running the GitHub Actions tests on your fork. While I don't expect any regression from this change, as it only alters comments, you should enable this for future backports at https://github.com/johnshajiang/jdk8u-dev/actions

gnu-andrew (approved):
Rather than treat this as a backport, it seems to make more sense to ensure that all occurrences of 'THL A29' are removed from the code base for each release.

As such, along with sanity checking the patch itself, I compared the list of files covered by the patch with those in the repository containing 'THL A29':

$ grep -r 'THL A29' * 2>&1 | tee /tmp/thl.8u
$ diffstat -p0 /tmp/8364597.17u > /tmp/diffstat.8u
$ for file in $(cut -d : -f 1 /tmp/thl.8u ); do if ! grep -q ${file} /tmp/diffstat.8u ; then echo "${file} missing";  fi; done

With this, I confirmed that this change covers all cases. The list was actually so short for this one, that it was clear from observation as well.

gnu-andrew:
Yes, these are known issues that we are working to resolve. I don't expect new failures from a comment change.

/approve yes

PR #746, opened 2026-01-16
8303215: Make thread stacks not use huge pages

tkiriyama:
Hi All,

I would like to backport the fix for bug JDK-8303215 to JDK 8. This bug addresses an issue with Transparent Huge Pages (THP) and thread stack allocation.
This fix is unclean, but almost clean backport from jdk11.

1. Backport:
The fix is almost clean from jdk11, but there is a slight difference.

- os_linux.cpp
Since JDK 8 does not include JDK-8183552, we need to add one more include statement. JDK-8183552 will likely not be backported to 8u.

- globals_linux_aarch64.hpp
In JDK 11, `CompilerThreadStackSize` is set to 2048, whereas JDK 8 originally sets it to `0`. This change to 2048 was introduced in JDK-8140520. Since JDK-8140520 is likely not directly related to this fix, I have excluded it from this fix and changed `CompilerThreadStackSize` to 2040.

2. Bug Reproduction:
The bug was successfully reproduced in Linux x86_64 with THP enabled (/sys/kernel/mm/transparent_hugepage/enabled = always).
Reproduction was confirmed using the following dummy thread program:
```
public class ThreadStackHugePageTest {
                final int NUM_THREADS = 1000;
                final long SLEEP_SECONDS = 3600;
                final long SLEEP_MILISECONDS = SLEEP_SECONDS * 1000;

                List<Thread> threads = new ArrayList<>();

                for (int i = 0; i < NUM_THREADS; i++) {
                        Runnable task = () -> {
                                try {
                                        Thread.sleep(SLEEP_MILISECONDS);
                                } catch (InterruptedException e) {
                                        Thread.currentThread().interrupt();
                                }
                        };

                        Thread thread = new Thread(task);
                        threads.add(thread);

                        thread.start();
                }

                System.out.println("all threads are started");

                for (Thread thread : threads) {
                        try {
                                thread.join();
                        } catch (InterruptedException e) {
                                Thread.currentThread().interrupt();
                        }
                }
        }
}
```
The fix significantly reduces the number of anonymous regions with fully occupied RSS.

3. Regression Testing:
I ran the test/hotspot tests on Linux x64 as a regression test. No failures were found.

Thank you.

---------
- [x] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).

---------
### Progress
- \[x\] Change must not contain extraneous whitespace
- \[x\] Commit message must refer to an issue
- \[x\] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))
- \[x\] [JDK-8303215](https://bugs.openjdk.org/browse/JDK-8303215) needs maintainer approval

### Issue
 * [JDK-8303215](https://bugs.openjdk.org/browse/JDK-8303215): Make thread stacks not use huge pages (**Enhancement** - P3 - Approved)

### Reviewers
 * [Paul Hohensee](https://openjdk.org/census#phh) (@phohensee - **Reviewer**) Review applies to [87b748cb](https://git.openjdk.org/jdk8u-dev/pull/746/files/87b748cb2c6983ec6740595884e18ca0bcc11401)
 * [Thomas Stuefe](https://openjdk.org/census#stuefe) (@tstuefe - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk8u-dev.git pull/746/head:pull/746` \
`$ git checkout pull/746`

`$ git checkout pull/746` \
`$ git pull https://git.openjdk.org/jdk8u-dev.git pull/746/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 746`

`$ git pr show -t 746`

</details>

<a href="https://git.openjdk.org/jdk8u-dev/pull/746.diff">https://git.openjdk.org/jdk8u-dev/pull/746.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jdk8u-dev/pull/746#issuecomment-3759779088)
</details>

### DIFF
--- a/hotspot/src/os_cpu/linux_aarch64/vm/globals_linux_aarch64.hpp
+++ b/hotspot/src/os_cpu/linux_aarch64/vm/globals_linux_aarch64.hpp
@@ -29,10 +29,18 @@
 // (see globals.hpp)
 
 define_pd_global(bool, DontYieldALot,            false);
-define_pd_global(intx, ThreadStackSize,          2048); // 0 => use system default
-define_pd_global(intx, VMThreadStackSize,        2048);
 
-define_pd_global(intx, CompilerThreadStackSize,  0);
+// Set default stack sizes < 2MB so as to prevent stacks from getting
+// large-page aligned and backed by THPs on systems where 2MB is the
+// default huge page size. For non-JavaThreads, glibc may add an additional
+// guard page to the total stack size, so to keep the default sizes same
+// for all the following flags, we set them to 2 pages less than 2MB. On
+// systems where 2MB is the default large page size, 4KB is most commonly
+// the regular page size.
+define_pd_global(intx, ThreadStackSize,          2040); // 0 => use system default
+define_pd_global(intx, VMThreadStackSize,        2040);
+
+define_pd_global(intx, CompilerThreadStackSize,  2040);
 
 define_pd_global(uintx,JVMInvokeMethodSlack,     8192);
 
--- a/hotspot/src/os_cpu/linux_aarch64/vm/globals_linux_aarch64.hpp
+++ b/hotspot/src/os_cpu/linux_aarch64/vm/globals_linux_aarch64.hpp
@@ -40,7 +40,7 @@ define_pd_global(bool, DontYieldALot,            false);
 define_pd_global(intx, ThreadStackSize,          2040); // 0 => use system default
 define_pd_global(intx, VMThreadStackSize,        2040);
 
-define_pd_global(intx, CompilerThreadStackSize,  2040);
+define_pd_global(intx, CompilerThreadStackSize,  0);
 
 define_pd_global(uintx,JVMInvokeMethodSlack,     8192);
 
--- a/hotspot/src/os/linux/vm/os_linux.cpp
+++ b/hotspot/src/os/linux/vm/os_linux.cpp
@@ -62,6 +62,7 @@
 #include "services/attachListener.hpp"
 #include "services/memTracker.hpp"
 #include "services/runtimeService.hpp"
+#include "utilities/align.hpp"
 #include "utilities/decoder.hpp"
 #include "utilities/defaultStream.hpp"
 #include "utilities/events.hpp"
@@ -901,6 +902,16 @@ bool os::create_thread(Thread* thread, ThreadType thr_type, size_t stack_size) {
     }
 
     stack_size = MAX2(stack_size, os::Linux::min_stack_allowed);
+
+    // Add an additional page to the stack size to reduce its chances of getting large page aligned
+    // so that the stack does not get backed by a transparent huge page.
+    size_t default_large_page_size = os::large_page_size();
+    if (default_large_page_size != 0 &&
+        stack_size >= default_large_page_size &&
+        is_aligned(stack_size, default_large_page_size)) {
+      stack_size += os::vm_page_size();
+    }
+
     pthread_attr_setstacksize(&attr, stack_size);
   } else {
     // let pthread_create() pick the default value.
@@ -3619,6 +3630,10 @@ bool os::Linux::setup_large_page_type(size_t page_size) {
 }
 
 void os::large_page_init() {
+  // Always initialize the default large page size even if large pages are not being used.
+  size_t large_page_size = Linux::setup_large_page_size();
+
+  // Handle the case where we do not want to use huge pages
   if (!UseLargePages &&
       !UseTransparentHugePages &&
       !UseHugeTLBFS &&
@@ -3636,7 +3651,6 @@ void os::large_page_init() {
     return;
   }
 
-  size_t large_page_size = Linux::setup_large_page_size();
   UseLargePages          = Linux::setup_large_page_type(large_page_size);
 
   set_coredump_filter();


phohensee (approved):

tstuefe:
@tkiriyama I will look at it in detail later. Where did you see the RSS problem? I mean OS version, kernel version, glibc version?

tkiriyama:
Hi @tstuefe ,
Thank you for checking.
I tested this in the following environment.

OS version - Red Hat Enterprise Linux 9.6 (Plow)
kernel version - 5.14.0-570.114.1.el9_6.x86_64
glibc version - glibc 2.34

hotspot/src/os_cpu/linux_aarch64/vm/globals_linux_aarch64.hpp
```diff
@@ -29,10 +29,18 @@
...
+// large-page aligned and backed by THPs on systems where 2MB is the
+// default huge page size. For non-JavaThreads, glibc may add an additional
+// guard page to the total stack size, so to keep the default sizes same
+// for all the following flags, we set them to 2 pages less than 2MB. On
+// systems where 2MB is the default large page size, 4KB is most commonly
+// the regular page size.
+define_pd_global(intx, ThreadStackSize,          2040); // 0 => use system default
+define_pd_global(intx, VMThreadStackSize,        2040);
+
+define_pd_global(intx, CompilerThreadStackSize,  2040);
```
tstuefe: No, pls revert the change to CompilerThreadStackSize. 

It usually defaults to VMThreadStackSize, which works. Unless ppc64 where we use 4MB stacks for some reason, but since THP page size is 16MB on ppc64, itmis not an issue.

Also, it does not matter anyway since there are typically very few compiler threads, so even if those form THPs that is not a big deal. The big deal are user-defined java threads, of which there can be thousands.

tstuefe (changes requested):

tstuefe:
Hi @tkiriyama,

I will continue reviewing this in the week after next; please refrain from downporting any more THP changes down to 8 for the time being, especially nothing to fix [JDK-8312182](https://bugs.openjdk.org/browse/JDK-8312182) - we have plans to fix this ourselves with a fix specifically targeted to JDK11 and 8.

tkiriyama:
@tstuefe 
Thank you for taking the time to look into the changes. I'm glad you're going to fix it.
Regarding this pull request, should I go ahead and close it, or should I keep working?
Also, I'm concerned that this fix has already been incorporated into JDK 11.

tstuefe:
@tkiriyama No, this change is fine; just revert the change to CompilerThreadStackSize and we are good to go from my side.

hotspot/src/os_cpu/linux_aarch64/vm/globals_linux_aarch64.hpp
tkiriyama: Understood. I fixed it, thanks!

tstuefe (approved):
Looks good. Thanks for fixing.

tstuefe:
Make sure you confirm adherence to the AI policy, otherwise jcheck will be unhappy.

tkiriyama:
@tstuefe ,
Sorry. After reviewing the policy, I added the result to my comment.

tstuefe:
@tkiriyama Please go ahead and apply for approval. Its a separate step needed for backport releases: first GH reviews, then approval separately from the maintainers specifically. You do that with the /approval backport command (see https://wiki.openjdk.org/spaces/SKARA/pages/56524908/Pull+Request+Commands)

Both me and @phohensee are reviewers in the jdk updates projects, but Paul's review is outdated. We need either another review or Paul needs to update his.

tkiriyama:
Hi @phohensee ,
Thank you for reviewing this PR earlier.
Following tstuefe's advice, some of the code has been modified. Although your status is currently "approved," could you please review it again?

gnu-andrew:
> @tkiriyama Please go ahead and apply for approval. Its a separate step needed for backport releases: first GH reviews, then approval separately from the maintainers specifically. You do that with the /approval backport command (see https://wiki.openjdk.org/spaces/SKARA/pages/56524908/Pull+Request+Commands)
> 
> Both me and @phohensee are reviewers in the jdk updates projects, but Paul's review is outdated. We need either another review or Paul needs to update his.

I can give it a review.  Even better if Paul gets to update his too.

gnu-andrew (approved):
The current version looks good and is close to the original patch.

I agree with Thomas' request to revert the `CompilerThreadStackSize` change.  I presume it was made in the original patch simply so that all three values stayed the same as each other.  By bumping it to `2040` in 8u, you would effectively be introducing part of 8140520 without any of the other changes to the stack size logic and adjustments to other architectures and operating systems.

[8140520](https://github.com/openjdk/jdk/commit/2bb3bc6449b0bf06324d52d2da89c5ce98eebab2) moves the handling of `CompilerThreadStackSize` for POSIX systems to `os_posix.cpp` and `get_initial_stack_size`.  It becomes used as a default size if no size is requested.  With the old logic in `os_linux.cpp`, `CompilerThreadStackSize` is used as is, so bumping it to 2040 would mean every compiler thread is now 2040.

This is the current 8u logic in `os_linux.cpp`:

~~~
      case os::compiler_thread:
        if (CompilerThreadStackSize > 0) {
          stack_size = (size_t)(CompilerThreadStackSize * K);
          break;
	}
~~~

In 11u, after 8140520, this code is replaced by `size_t stack_size = os::Posix::get_initial_stack_size(thr_type, req_stack_size);` with `req_stack_size` being a new parameter passed to `os::create_thread`.  The logic for `os::compiler_thread` in `get_initial_stack_size` is:

~~~
  case os::compiler_thread:
    if (req_stack_size == 0 && CompilerThreadStackSize > 0) {
      // no requested size and we have a more specific default value
      stack_size = (size_t)(CompilerThreadStackSize * K);
    }
    stack_size = MAX2(stack_size,
                      _compiler_thread_min_stack_allowed);
    break;
~~~

so `stack_size` is no longer always set to a positive `CompilerThreadStackSize` value.  Adding the size change to 8u without the logic change would result in different behaviour to that in 11u.  So thanks for dropping that part of the patch.

tstuefe:
@tkiriyama can you do the apply-for-approval step? I would like for this patch to go in, since its the base for my coming JDK 8 THP fixes.

tkiriyama:
/approval request I would like to backport this bug (JDK-8303215) done in jdk17 to jdk11. Even though it is not a clean backport,  it has been sufficiently reviewed.

tstuefe:
@tkiriyama Can you investigate the crash in the linux hotspot tests? https://github.com/tkiriyama/jdk8u-dev/actions/runs/27738928666/job/82063041225

gnu-andrew:
> @tkiriyama Can you investigate the crash in the linux hotspot tests? https://github.com/tkiriyama/jdk8u-dev/actions/runs/27738928666/job/82063041225

It's worth double-checking but I think this may be one of a few GC tests that sporadically fails on many 8u PRs. For instance, a couple failed on [this GHA PR](https://github.com/zzambers/jdk8u-dev/actions/runs/25750392312/job/75631070256) which only updates the name of a package dependency in the GitHub actions.  I couldn't find a PR with that exact test, but I feel I've seen it before.

tkiriyama:
@tstuefe , @gnu-andrew ,
Thanks for your comments.
I reran the GHA, and the problem test passed. To be honest, I'm not sure of the exact cause, but it doesn't seem to me like this fix had any impact.
Can I proceed to the integration?

gnu-andrew:
Passed locally for me too.  As I said above, I believe it's sporadic.

/approve yes

